PR #4, opened 2020-10-31
Reverted OpenGL to 330, regenerated glad and changed shaders. If this…

UbiBelETF:
… works for everyone, it can stay in the main branch too!

I'm not sure it will work for everyone, however. 
It would be great if people would update their drivers before trying this out. So, update your NVidia/AMD/Intel drivers to the newest you can find, then go to the root folder of the repository and type:

```git checkout engine/opengl330_fallback```

Once that's done, generate the projects again via the script, go into Visual Studio, build and run, and tell me what you've got.

Expectations:

- the `WARNING: 0:5: '#extension' :  'GL_ARB_gpu_shader_int64' is not supported` message will now be gone.
- any sort of OpenGL error that has to do with 4.0+ versions will be gone.
- `bindless_textures_extension` might still be missing, but maybe not!

Write your findings in the comments below.


Yugo0:
I updated the drivers, and when running the project in main branch the problem I had up to that point was replaced by the one some of the other colleagues had, i.e. assretion at line 96 of the texture.h file.

After switching to engine/opengl330_fallback branch, the problem persisted.

gavrilovmiroslav:
And if you remove the assertion in that line?

Yugo0:
Yes, that seems to work, game is running without any issues, as far as I can tell. I did the exact same thing in the main branch (same commit that we used on Saturday) and it worked there as well.

PR #8, opened 2020-11-03
Unix Support

harry-cpp:


source/dagger/core/optimization.h
```diff
@@ -2,9 +2,5 @@
 
 #ifndef USE_DEDICATED_GPU
 #define USE_DEDICATED_GPU
-extern "C"
```
harry-cpp, line 5: Need to add define here.

.vscode/tasks.json
```diff
@@ -0,0 +1,13 @@
+{
+	"version": "2.0.0",
+	"tasks": [
+		{
```
harry-cpp, line 4: Fix formatting.

UbiBelETF:
Admitted to `main` for easier work on both MacOS and Linux versions.

PR #14, opened 2020-11-05
Engine/weekly upgrade

gavrilovmiroslav:
Camera correction: because of this, ping pong scale broke completely. Look at `Camera` to see how to translate between window and world.


data/shaders/standard.fs.glsl
```diff
@@ -9,5 +9,8 @@ out vec4 o_FragColor;
 
 void main()
 {
-	o_FragColor = texture(u_Texture, v_TextureCoord) * v_QuadColor;
+	
+	vec4 tex = texture(u_Texture, v_TextureCoord);
+	if(tex.a < 0.99) discard;
```
DPrivalov, line 14: do we need this discard?

source/dagger/gameplay/ping_pong/ping_pong_main.h
```diff
@@ -2,12 +2,19 @@
 
 #include "core/core.h"
 #include "core/system.h"
+#include "core/game.h"
+#include "core/engine.h"
 
 using namespace dagger;
 
-namespace pingPong
```
DPrivalov, line 8: I think we should leave namespace here, it'll be easier with code complete

source/dagger/gameplay/platformer/platformer_controller.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+#include "core/core.h"
+#include "core/system.h"
+
+using namespace dagger;
+
+class PlatformerControllerSystem
```
DPrivalov: namespace here will be good too

DPrivalov (changes requested):

data/shaders/standard.fs.glsl
gavrilovmiroslav: Sadly, for now. I'm looking into it, and it might be an issue with geometry instantiation (I'd check without it but it's quite a rewrite, so I'm going for speed of delivery here), but it should be noted that it's not here to stay, but also is a fully valid shader language thing you shouldn't really look badly upon :P

source/dagger/gameplay/ping_pong/ping_pong_main.h
gavrilovmiroslav: Do you think `snake_case` looks better for namespaces? I kinda do, maybe I'd change that if it's all the same to you.

source/dagger/gameplay/platformer/platformer_controller.h
gavrilovmiroslav: 👍

DPrivalov (approved):

PR #20, opened 2020-11-08
todo in pongpong was fixed

DPrivalov:


source/dagger/gameplay/ping_pong/player_scores.cpp
```diff
@@ -49,7 +49,7 @@ void PlayerScoresSystem::Run()
 
     if (ballOnField == 0)
     {
-        // TODO: set speed to be random in both directions
-        CreatePingPongBall(s_TileSize, ColorRGBA(1, 1, 1, 1), { rand()%10 + 4,rand()%10 + 4,0 },   { 0,rand()%(s_FieldHeight / 2),0 });
+        float xSpeed = rand() % 10 + 4;
+        CreatePingPongBall(s_TileSize, ColorRGBA(1, 1, 1, 1), { xSpeed ,rand()%10 + 4,0 },   { 0,rand()%(s_FieldHeight / 2),0 });
```
UbiBelETF, line 53: This still doesn't work: it should be able to send the ball to the left as well. Please fix! <3

UbiBelETF (changes requested):
Please fix this so that the right player isn't always targetted first.

PR #21, opened 2020-11-08
added negative direction speed

nkristina:


source/dagger/gameplay/ping_pong/player_scores.cpp
```diff
@@ -50,6 +50,7 @@ void PlayerScoresSystem::Run()
     if (ballOnField == 0)
     {
         // TODO: set speed to be random in both directions
-        CreatePingPongBall(s_TileSize, ColorRGBA(1, 1, 1, 1), { rand()%10 + 4,rand()%10 + 4,0 },   { 0,rand()%(s_FieldHeight / 2),0 });
+		float x_speed = rand() % 10 - 3.12;
```
DPrivalov: x_speed can be generated with 0, so the game will freeze, please try to fix it

DPrivalov (changes requested):

source/dagger/gameplay/ping_pong/player_scores.cpp
UbiBelETF: Also, `3.12` is a double and not a float (like `3.12f` is). Watch out for type sizes and implicit conversions...

milosmilosevic13 (approved):

source/dagger/gameplay/ping_pong/player_scores.h
```diff
@@ -28,5 +28,12 @@ namespace ping_pong
             s_FieldHeight = height;
             s_TileSize = tileSize;
         }
+
+		float GetRandomSpeed()
+		{
+			float x_speed = rand() % 10 - 5.0f;
+			x_speed = x_speed ? x_speed : x_speed + 5.0f;
```
DPrivalov: you are trying to compare with 0 here?
it's bad practice to compare float with 0 like that. You need to use > < operators instead.

DPrivalov (changes requested):

DPrivalov (approved):
ball still might be too slow, but it's much better already

PR #22, opened 2020-11-08
Team/fried cucumber/fix/todo pinpong fixed

Nikolar97:


source/dagger/gameplay/ping_pong/player_scores.cpp
```diff
@@ -49,7 +49,6 @@ void PlayerScoresSystem::Run()
 
     if (ballOnField == 0)
     {
-        // TODO: set speed to be random in both directions
-        CreatePingPongBall(s_TileSize, ColorRGBA(1, 1, 1, 1), { rand()%10 + 4,rand()%10 + 4,0 },   { 0,rand()%(s_FieldHeight / 2),0 });
+        CreatePingPongBall(s_TileSize, ColorRGBA(1, 1, 1, 1), { (1 - 2 * (rand() % 2)) * (rand()%10 + 4),(rand()%10 + 4)*(1 - 2 * (rand() % 2)),0 },   { 0,rand()%(s_FieldHeight / 2),0 });
```
UbiBelETF: Hmmm, I have to think about what this does. Can you maybe write it in a more readable way? Maybe separate it into multiple variables beforehand, or write a function so that we could reuse it?

Nikolar97: The code is more readable now.

Nikolar97 (commented):
Should be much more readable now

DPrivalov (approved):

DPrivalov:
great one!

PR #26, opened 2020-11-08
Team/sun tsu/fix/todo pingpong fixed

uros-maricic:


DPrivalov (approved):
Good!

source/dagger/gameplay/ping_pong/player_scores.cpp
```diff
@@ -50,6 +50,12 @@ void PlayerScoresSystem::Run()
     if (ballOnField == 0)
     {
         // TODO: set speed to be random in both directions
-        CreatePingPongBall(s_TileSize, ColorRGBA(1, 1, 1, 1), { rand()%10 + 4,rand()%10 + 4,0 },   { 0,rand()%(s_FieldHeight / 2),0 });
+        float xSpeed = rand() % 17 - 8;
+        while ((xSpeed > -3) && (xSpeed < 3))
+        {
+            xSpeed = rand() % 17 - 8;
+        }
```
UbiBelETF, line 57: You can use `std::clamp` instead. https://en.cppreference.com/w/cpp/algorithm/clamp

UbiBelETF: So basically, `std::clamp(rand() % 17 - 8, -3, 3)` should do the trick.

gavrilovmiroslav: Also, take into account what happens if your value here ends up being 0...

uros-maricic: Alright, thank you for reviewing and for the helpful suggestion. Cheers!

PR #30, opened 2020-11-08
todo in pingpong was fixed

kostaDimitrijevic:


source/dagger/gameplay/ping_pong/player_scores.cpp
```diff
@@ -11,7 +11,7 @@ using namespace ping_pong;
 int PlayerScoresSystem::s_FieldWidth = 20;
 int PlayerScoresSystem::s_FieldHeight = 10;
 Float32 PlayerScoresSystem::s_TileSize = 1.f;
-
+bool k = true;
```
UbiBelETF: Please use more descriptive variable names. Names get stripped by the compiler anyway, and it helps everyone to know what `k` does.

smiljanaspasic (approved):

source/dagger/gameplay/ping_pong/player_scores.cpp
```diff
@@ -46,10 +46,17 @@ void PlayerScoresSystem::Run()
...
         }
     }
-
     if (ballOnField == 0)
-    {
+    {   
+        
         // TODO: set speed to be random in both directions
-        CreatePingPongBall(s_TileSize, ColorRGBA(1, 1, 1, 1), { rand()%10 + 4,rand()%10 + 4,0 },   { 0,rand()%(s_FieldHeight / 2),0 });
+        if (k) {
```
UbiBelETF: Is this meant to balance the game so that one ball goes left, and then the next goes right?

Yugo0 (approved):
Maybe in future we change it to go random to one side or the other. Just a suggestion.

source/dagger/gameplay/ping_pong/player_scores.cpp
```diff
@@ -46,10 +46,17 @@ void PlayerScoresSystem::Run()
...
+            CreatePingPongBall(s_TileSize, ColorRGBA(1, 1, 1, 1), { rand() % 10 + 5, rand() % 10 + 4,0 }, { 0,rand() % (s_FieldHeight / 2),0 });
+            k = false;
+        }
+        else {
+            CreatePingPongBall(s_TileSize, ColorRGBA(1, 1, 1, 1), { rand() % 10 - 15, rand() % 10 + 4,0 }, { 0,rand() % (s_FieldHeight / 2),0 });
+            k = true;
```
UbiBelETF: This would be better as a single `CreatePingPongBall`, with a variable called `offset` for example, that will be `5` or `-15` added to `rand()`. After that you can say `k = !k;` with whatever name you choose for `k` and lower repetition.

UbiBelETF (changes requested):

source/dagger/gameplay/ping_pong/player_scores.cpp
kostaDimitrijevic: yes

UbiBelETF: Maybe go with `nextBallDirectionIsLeft` or something like that

Yugo0 (approved):

Yugo0 (approved):

Yugo0 (approved):

smiljanaspasic (commented):
In this fix I tried to make it more random

smiljanaspasic:
> 
> 
> or rather, if the sum with offset ends up being zero. Also, why the (5 + 1) + 15 sum?

> 
> 
> what happens if `offset` ends up being zero?

So yeah that could be a problem,I could change it so that if it ends up being zero I just change it to one
As for the other question I tried to make the offset take values in range from -15 to 5 so I used this formula:  min+rand()%((max+1)-min) where min= -15 and max=5  so in the formula we have -15++rand()%((5+1)-(-15)) which is equal to 15++rand()%((5+1)+15)

gavrilovmiroslav:
> > or rather, if the sum with offset ends up being zero. Also, why the (5 + 1) + 15 sum?
> 
> > what happens if `offset` ends up being zero?
> 
> So yeah that could be a problem,I could change it so that if it ends up being zero I just change it to one
> As for the other question I tried to make the offset take values in range from -15 to 5 so I used this formula: min+rand()%((max+1)-min) where min= -15 and max=5 so in the formula we have -15++rand()%((5+1)-(-15)) which is equal to 15++rand()%((5+1)+15)

What if we skip both problems by splitting the problem in two: find a way to generate the sign (-1 or 1) and then multiply that by a positive number in a range you like! :)

PR #32, opened 2020-11-08
Added other team members. Made code fix more readable.

senadkurtisi:
Ball is a bit faster now on the x axis.

### DIFF
--- a/source/dagger/gameplay/ping_pong/player_scores.cpp
+++ b/source/dagger/gameplay/ping_pong/player_scores.cpp
@@ -49,6 +49,8 @@ void PlayerScoresSystem::Run()
 
     if (ballOnField == 0)
     {
-        CreatePingPongBall(s_TileSize, ColorRGBA(1, 1, 1, 1), { rand() % 10 - 5,rand() % 10 + 4,0 }, { 0,rand() % (s_FieldHeight / 2),0 });
+        CreatePingPongBall(s_TileSize, ColorRGBA(1, 1, 1, 1), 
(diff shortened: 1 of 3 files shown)


source/dagger/gameplay/ping_pong/player_scores.cpp
```diff
@@ -49,6 +49,8 @@ void PlayerScoresSystem::Run()
 
     if (ballOnField == 0)
     {
-        CreatePingPongBall(s_TileSize, ColorRGBA(1, 1, 1, 1), { rand() % 10 - 5,rand() % 10 + 4,0 }, { 0,rand() % (s_FieldHeight / 2),0 });
+        CreatePingPongBall(s_TileSize, ColorRGBA(1, 1, 1, 1), 
+                           { rand() % 20 - 10, rand() % 10 + 4, 0 },
```
UbiBelETF: What happens if `rand()` returns `5`?

senadkurtisi: In that case you get -5, because subtracting 10 from the rand() % 20 centers the range around 0. The original range is [0, 20), and now it is [-10, 10) if I got that right.

UbiBelETF: My bad, what if `rand()` produces `10`?

UbiBelETF: (I was looking at the old code and commenting before)

PR #36, opened 2020-11-10
Added new component and new system, players now flash with red color after collision

Yugo0:


DPrivalov (approved):
nice first system! gj!

source/dagger/gameplay/ping_pong/impact_flash.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+#include <core/system.h>
+
+using namespace dagger;
+
+struct ImpactFlash
+{
+    float flashingRate{ 4.0f };
```
gavrilovmiroslav: Look at core.h for the type aliases we should start using slowly. In short, we have aliases for all base types (and some complex ones) which should be used because they abstract away the implementation beneath. In this case, for example, it would be better (albeit semantically equivalent) to use `Float32`.

kostaDimitrijevic (approved):

smiljanaspasic (approved):

PR #38, opened 2020-11-11
Added power up feature. Players speed up and flicker after defending

senadkurtisi:
Flickering is just an indicator for the player/s that their power-up is still active.

### DIFF
--- a/source/dagger/gameplay/ping_pong/power_up.h
+++ b/source/dagger/gameplay/ping_pong/power_up.h
@@ -4,26 +4,29 @@
 
 using namespace dagger;
 
-struct PowerUp
+namespace ping_pong
 {
-	Bool is_active{ false };
+	struct PowerUp
+	{
+		Bool is_active{ false };
 
-	Float32 speed_multiplier{ 2.5f };
+		Float32 speed_multiplier{ 2.5f };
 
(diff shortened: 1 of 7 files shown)


source/dagger/gameplay/ping_pong/power_up.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+#include "core/system.h"
+#include "core/core.h"
+
+using namespace dagger;
+
+struct PowerUp
```
DPrivalov: place it inside ping_pong namespace as any other pingpong system

DPrivalov (changes requested):
Good job! 

One important thing to change and it's good to merge!

DPrivalov (approved):

UrosBorovcanin (commented):
The code functions as it should and functions well. The only negative side that i see here is that it doesn't conform to the styleguide.md

DPrivalov (approved):

UrosBorovcanin (approved):
The code is up to the style guide standard and still works completely fine.

DPrivalov:
@Djafex, waiting for u =)

FedjaVujicic (approved):

PR #39, opened 2020-11-12
Added a slow down time power up

Coakim:
Added a power up which slows down time .
Each player has the power up bound to a different key bind ( E - for player one , L- for player 2). The power up slows down all of the balls currently on the field by half , and both of the walls controlled by the players by a third. Each player can activate the power up 2 times per game , and the power up can be deactivated by pressing the same button while the power up is active. Also , power up deativates automatically after 5 seconds if it isn't deactivated manually.


source/dagger/gameplay/ping_pong/pingpong_playerinput.cpp
```diff
@@ -1,5 +1,5 @@
 #include "pingpong_playerinput.h"
-
+#include "pingpong_ball.h"
```
DPrivalov: use full path for files different from this one

source/dagger/gameplay/ping_pong/pingpong_playerinput.cpp
```diff
@@ -23,34 +28,107 @@ void PingPongPlayerInputSystem::WindDown()
...
-        else if (kEvent_.key == ctrl_.down_key && (kEvent_.action == EDaggerInputState::Held || kEvent_.action == EDaggerInputState::Pressed))
-        {
-            ctrl_.input.y = -1;
-        }
-        else if (kEvent_.key == ctrl_.down_key && kEvent_.action == EDaggerInputState::Released && ctrl_.input.y < 0)
-        {
-            ctrl_.input.y = 0;
-        }
-    });
+        auto& ctrl_ = view.get<ControllerMapping>(entity);
```
DPrivalov: we use _ in the end of the name for parameters. for example kEvent_. 
local variables should not have it.

source/dagger/gameplay/ping_pong/pingpong_playerinput.cpp
```diff
@@ -23,34 +28,107 @@ void PingPongPlayerInputSystem::WindDown()
...
+                s_ActivatePowerUp = false;
+
+                Engine::Registry().view<PingPongBall>().each([&](PingPongBall& ball) {
+
+                    if (ball.reachedGoal)
+                    {
+                        return;
+                    }
+
+                    ball.speed.x = ball.speed.x * 0.33f;
```
DPrivalov: it'll be good to store 0.33f as a const somewhere, so it'll be easier to change it later.

source/dagger/gameplay/ping_pong/player_scores.cpp
```diff
@@ -49,7 +50,13 @@ void PlayerScoresSystem::Run()
 
     if (ballOnField == 0)
     {
-        // TODO: set speed to be random in both directions
-        CreatePingPongBall(s_TileSize, ColorRGBA(1, 1, 1, 1), { rand()%10 + 4,rand()%10 + 4,0 },   { 0,rand()%(s_FieldHeight / 2),0 });
+        //If the new ball is to be created when the slow-mode power up is active , it is created with the slowed down speed
```
DPrivalov: it's better to use new field inside ball, something like speed multiplier. And you will be setting it to 1 or 0.33f

DPrivalov (changes requested):

source/dagger/gameplay/ping_pong/player_scores.cpp
Coakim: Added a field to the PingPongBall component , so it isn't necessary to change ball speed every time

source/dagger/gameplay/ping_pong/pingpong_playerinput.cpp
Coakim: Added two consts , for ball and player speed multipliers, as you suggested

source/dagger/gameplay/ping_pong/pingpong_playerinput.cpp
Coakim: Sorry about that , fixed it now

source/dagger/gameplay/ping_pong/pingpong_playerinput.cpp
Coakim: Changed the path to full path

DPrivalov (approved):
now it's nice, ask your team to try it out and approve

DamjanTM (approved):

PR #34, opened 2020-11-10
Ball speeds up every 5 seconds

Yugo0:

PR #40, opened 2020-11-12
Added a malicious ball to the game

Coakim:
Added a chance (10%) that ball becomes malicious upon creation.
The malicious ball looks the same as the normal ball ,but if it colides with a player wall it inverts the controls of that player for 5 seconds (up_key makes the wall go down , and vice versa) and gets destroyed immediately.


source/dagger/gameplay/ping_pong/pingpong_playerinput.h
```diff
@@ -11,6 +11,8 @@ namespace ping_pong
     {
         EDaggerKeyboard up_key;
         EDaggerKeyboard down_key;
+        int inverted{1};
+        TimePoint inverted_time_point{};
```
DPrivalov: use Float32 timer instead of TimePoint, example is here:
https://github.com/UbiBelETF/dagger/pull/34/files

source/dagger/gameplay/ping_pong/pingpong_ball.cpp
```diff
@@ -60,7 +60,15 @@ void PingPongBallSystem::Run()
                 if (std::abs(collisionSides.y) > 0)
                 {
                     ball.speed.y *= -1;
-                }
+                }  
+            }
+
+            if (Engine::Registry().has<ControllerMapping>(col.colidedWith) && ball.isMalicious == true) {
```
DPrivalov: it's better to check light things first:
if (ball.isMalicious == true && Engine::Registry().has<ControllerMapping>(col.colidedWith)) {

if ball.isMalicious isn't true then you wouldn't need to  check further things, and c++ is doing this for you.

DPrivalov (changes requested):

source/dagger/gameplay/ping_pong/pingpong_playerinput.h
Coakim: Thanks for the advice , I'll change it asap

source/dagger/gameplay/ping_pong/pingpong_ball.cpp
Coakim: Oh , yeah... You're right. I rushed it a bit and didn't notice . Will fix that too

source/dagger/gameplay/ping_pong/pingpong_playerinput.h
Coakim: Used Float32 instead of TimePoint as you mentioned , should be ok regarding that part now

source/dagger/gameplay/ping_pong/pingpong_ball.cpp
Coakim: Swapped the conditions as you sugested

source/dagger/gameplay/ping_pong/pingpong_ball.cpp
```diff
@@ -60,7 +60,14 @@ void PingPongBallSystem::Run()
...
                 {
                     ball.speed.y *= -1;
-                }
+                }  
+            }
+
+            if (ball.isMalicious == true &&  Engine::Registry().has<ControllerMapping>(col.colidedWith)  ) {
+               auto &cntrl =  Engine::Registry().get<ControllerMapping>(col.colidedWith);
+               cntrl.inverted = -1;
+               Engine::Registry().destroy(entity);
```
DPrivalov: @gavrilovmiroslav , what do u think, should we do that in EndOfFrame or it's fine here?

DPrivalov (approved):

source/dagger/gameplay/ping_pong/pingpong_ball.cpp
gavrilovmiroslav: I had a case where it actually made a difference, so I'd play it safe and delegate to EndOfFrame. Maybe we can even make a small helper mechanism for this behavior sometime later so that we don't have to always do this ourselves...

DPrivalov: @Coakim , put some flag into ball instead of deleting it and then in EndOfFrame function delete it.

You can find an example how to subscribe to EndFrame event here:
RacingCollisionsLogicSystem::OnEndOfFrame

Coakim: I was just looking for a way to destroy a specific ball and found the Engine::Registry().destroy function , but I guess I didn't think about whether it could cause problems doing it mid frame. Will move the deletion to EndOfFrame function as suggested

Coakim: Moved the deletion to EndOfFrame

DamjanTM (approved):

PR #43, opened 2020-11-13
Added laser feature to the racing game, along with a simple objective

UrosBorovcanin:


DPrivalov (approved):
Nice one!

senadkurtisi (commented):
This feature is excellent and the code runs efficiently. 
Maybe it would be a good idea to wait for something like 0.1s after destroying the target number of cars and then restart the game? While playing I thought that I have collided with the enemy car and the game restarted because of it, but it turned out I won. Nevertheless I definitely like it

senadkurtisi (approved):
Awesome!

FedjaVujicic (approved):

PR #56, opened 2020-11-15
Added the roll movement for both players in the platformer

senadkurtisi:
Players can now use the roll movement. WASD player rolls using the X key, while Arrows player rolls using the right ctrl key.

### DIFF
--- a/source/dagger/gameplay/platformer/platformer_controller.cpp
+++ b/source/dagger/gameplay/platformer/platformer_controller.cpp
@@ -44,13 +44,6 @@ void PlatformerControllerSystem::Run()
                     sprite_.scale.x = run;
                 }
             }
-            else
-            {
-                if (char_.timeRolling >= char_.rollingTime)
-                {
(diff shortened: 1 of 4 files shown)


source/dagger/gameplay/platformer/platformer_controller.cpp
```diff
@@ -27,18 +27,50 @@ void PlatformerControllerSystem::SpinUp()
...
+                // allow him to do that. Alligning the sprite_.scale.x with run variable enables
+                // the character to start rolling in the direction it is facing
+                if (run != 0)
+                {
+                    sprite_.scale.x = run;
+                }
+
+                if (char_.isRolling)
+                {
+                    if (char_.timeRolling < char_.rollingTime)
```
DPrivalov: so it's just rolling for 5 frames? why not to make it Float32 time dependent? 0.5 sec or something?

FedjaVujicic (approved):
Works great

PR #57, opened 2020-11-17
Adding roll, attack,block and drink potion options for the playable c…

SofijaBosk:


source/dagger/gameplay/platformer/platformer_controller.cpp
```diff
@@ -27,19 +27,50 @@ void PlatformerControllerSystem::SpinUp()
...
                 AnimatorPlay(animator_, "souls_like_knight_character:RUN");
                 sprite_.scale.x = run;
                 sprite_.position.x += char_.speed * sprite_.scale.x * Engine::DeltaTime();
             }
+
+            if(run==0 && useP==0 && down==0 && block==0 && sAttack==0) 
+            {
+                AnimatorPlay(animator_, "souls_like_knight_character:IDLE");
+            }
+
```
UbiBelETF, line 98: Here's a thought: could I trick this system into enabling me to cancel a block/attack/jump half way and begin another animation?

SofijaBosk: It could probably, but I must admit I'm not sure how exactly to implement it. Maybe adding boolean statements in PlatformerCharacter for every action and make it false if the action we want to have a higher priority activates?

PR #58, opened 2020-11-17
Team/delta snakes/feature/jump_movement

FedjaVujicic:
Added a jump feature to the platformer game. Changing a direction while rolling is now possible without pressing the roll command.

### DIFF
--- a/source/dagger/gameplay/platformer/platformer_controller.cpp
+++ b/source/dagger/gameplay/platformer/platformer_controller.cpp
@@ -29,6 +29,17 @@ Float32 platformer::CalculateVerticalSpeed(Float32 speed_, Float32 acceleration_
     return speed_ - acceleration_ * time_;
 }
 
+// Ex: If the player is pressing the left movement key and roll Key
+// allow him to do that. Alligning the sprite_.scale.x with run variable enables
+// the character to start rolling in the direction it is facing
+void platformer::AlignCharacter(Float32 run_, Sprite& sprite_)
+{
+    if (run_ != 0)
+    {
+        sprite_.scale.x = run_;
+    }
+}
+
 void PlatformerControllerSystem::Run()
 {
     Engine::Registry().view<InputReceiver, Sprite, Animator, PlatformerCharacter>().each(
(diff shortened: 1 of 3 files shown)


source/dagger/gameplay/platformer/platformer_controller.cpp
```diff
@@ -45,25 +53,75 @@ void PlatformerControllerSystem::Run()
...
+                        sprite_.scale.x = run;
+                    }
+                    Float32 speed = CalculateVerticalSpeed(char_.verticalInitialSpeed, char_.gravity, char_.timeJumping);
+                    if (!char_.reachedMax)
+                    {
+                        AnimatorPlay(animator_, "souls_like_knight_character:JUMP");
+                        char_.timeJumping += Engine::DeltaTime();
+                        sprite_.position.y += speed * Engine::DeltaTime();
+                        if (char_.isRunning) sprite_.position.x += char_.speed * sprite_.scale.x * Engine::DeltaTime();
+                        else sprite_.position.x += sprite_.scale.x * Engine::DeltaTime();
```
senadkurtisi: I would add a set of curly braces every time we use if/else in order to maintain readability. 
And maybe use it as (this second part is just my preference):
if (condition)
{
}
instead of:
if (condition) {
}

source/dagger/gameplay/platformer/platformer_controller.cpp
```diff
@@ -45,25 +53,75 @@ void PlatformerControllerSystem::Run()
...
-                    char_.isRolling = false;
-                    char_.timeRolling = 0;
+                    if (run != 0)
+                    {
+                        sprite_.scale.x = run;
+                    }
+                    Float32 speed = CalculateVerticalSpeed(char_.verticalInitialSpeed, char_.gravity, char_.timeJumping);
+                    if (!char_.reachedMax)
+                    {
+                        AnimatorPlay(animator_, "souls_like_knight_character:JUMP");
```
senadkurtisi, line 101: Maybe we can add lines 95-99 in some function considering the fact that similar lines appear in the 105-109? And based on the direction of the movement we could play different animation?

senadkurtisi (commented):
It works really well and smooth! I have small suggestions based on the style of the code, but besides that it is really good!

source/dagger/gameplay/platformer/platformer_controller.cpp
UbiBelETF: This is the part of the styleguide I need to add, still. If in a situation to use `{}`, try to always use them (we sometimes leave them and then review and get them in later), but always put them in the next line. It's very important for readability. Next year, we'll maybe even have a linter for all of this, to give you useful comments and auto-corrections on the spot... not this one, though :D

source/dagger/gameplay/platformer/platformer_controller.cpp
FedjaVujicic: Yes. I actually intended to do this, but forgot :laughing:

senadkurtisi (approved):
Great!

UrosBorovcanin (approved):
The jumping feels a little floaty to me, but thats something thats easily tunable by editing gravity and intial speed of the character, and is not that important at this stage. Otherwise i'd say that these additions are well thought out, well written and they work well.

PR #59, opened 2020-11-18
Added option to have circular hitboxes

milosmilosevic13:


milosmilosevic13:
I'm not sure if the GetCollisionSides and GetCollisionCenter are affected by this change, so I left them unchanged for now.

source/dagger/gameplay/common/simple_collisions.cpp
```diff
@@ -3,6 +3,9 @@
 #include "core/engine.h"
 #include "core/game/transforms.h"
 
+#include <iostream>
```
DPrivalov, line 6: do u need this?

DPrivalov (approved):
try to split into more smaller pull requests next time

source/dagger/gameplay/common/simple_collisions.cpp
milosmilosevic13: No, it was used for some debugging but wasn't removed afterwards, my bad.

nkristina (approved):

mmacura9 (approved):

PR #61, opened 2020-11-18
Added the pause game feature

senadkurtisi:
Added ping_pong_main.cpp to illustrate the usage of this feature.

### DIFF
--- a/source/dagger/core/engine.cpp
+++ b/source/dagger/core/engine.cpp
@@ -107,7 +107,7 @@ void Engine::EngineStop()
 	this->m_Registry.reset();
 }
 
-void Engine::ToggleSystems(Bool toPause_)
+void Engine::ToggleSystemsPause(Bool toPause_)
 {
 	s_Instance->s_IsPaused = toPause_;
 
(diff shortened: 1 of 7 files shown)


source/dagger/core/engine.cpp
```diff
@@ -103,3 +106,24 @@ void Engine::EngineStop()
 	this->m_EventDispatcher.reset();
 	this->m_Registry.reset();
 }
+
+void Engine::ToggleSystems(Bool toPause_)
```
DPrivalov: ToggleSystemsPause

source/dagger/core/system.h
```diff
@@ -14,6 +14,15 @@ namespace dagger
 		virtual inline void SpinUp() {};
 		virtual inline void Run() {};
 		virtual inline void WindDown() {};
+
+		Bool canBePaused{ false };
+		Bool isPaused{ false };
+
+		System() {}
+		System(Bool canBePaused_): canBePaused(canBePaused_) {}
```
DPrivalov: seems like this one is not used

DPrivalov (changes requested):

DPrivalov (approved):
ok, let's wait for @UbiBelETF to review

PR #65, opened 2020-11-20
Adding the configure imgui window and editing the ping pong game so t…

Nikolar97:
Added a new imgui window to the ping pong game which allows adjusting some parameters and pausing the game

### DIFF
--- a/source/dagger/gameplay/ping_pong/pingpong_ball.cpp
+++ b/source/dagger/gameplay/ping_pong/pingpong_ball.cpp
@@ -5,7 +5,7 @@
 #include "core/graphics/sprite.h"
 
 #include "gameplay/common/simple_collisions.h"
-
+#include "gameplay/ping_pong/configure.h"
 #include <algorithm>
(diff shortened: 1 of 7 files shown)


source/dagger/gameplay/ping_pong/pingpong_ball.cpp
```diff
@@ -74,7 +74,7 @@ void PingPongBallSystem::Run()
         }
         else
         {
-            t.position += (ball.speed * Engine::DeltaTime());
+            t.position += (ball.speed * Engine::DeltaTime()*ConfigureSystem::BallSpeedMultiplier()*ConfigureSystem::GameSpeedMultiplier());
```
DPrivalov, line 77: it's kinda debug system, but you are using it without #if defined(DAGGER_DEBUG)

I suggest you to move these BallSpeedMultiplier and GameSpeedMultiplier to some none debug(non imgui) class, and make tools(imgui) change them

PR #69, opened 2020-11-20
Team/morning afternoon/feature/platformer movement and combat system

Coakim:
Added more movement animations to the platformer game , as well as some attack animations (jumping , rolling , heavy , light ,  down attack). Players can use jump while in the air to perform a double jump. Rolling is performed by tapping the same key used for running 2 times fast, which performs a roll in the selected direction.Implemented a simple combat system which tracks both players health points (max 100) , and the health bars for both players are displayed on top of the screen. Used the existing collision system to detect collision , and if the collision is detected while an attack is performed the players on the other end of the attack takes damage based on the type of the attack performed. The damage is visibly displayed as the health bar drains.
If the player reaches 0 health he dies , making that player unable to recieve any input (Didn't implement stopping the game when the player dies yet).
*Was not able to make the health bars follow the camera properly (It is possible for health bars to dissapear from the view).Tried using the parallax system but failed ...


source/dagger/gameplay/platformer/platformer_combat.cpp
```diff
@@ -0,0 +1,71 @@
...
+	for (auto entity : view) {
+		SimpleCollision& col = Engine::Registry().get<SimpleCollision>(entity);
+		Transform& t = Engine::Registry().get<Transform>(entity);
+		PlatformerCharacter& character = Engine::Registry().get<PlatformerCharacter>(entity);
+		CharacterHealth& chealth = Engine::Registry().get<CharacterHealth>(entity);
+
+		if (character.dead) {
+			return;
+		}
+		if (col.colided) {
```
DPrivalov: might be good to check if character in any attack state before going into collision processing

source/dagger/gameplay/platformer/platformer_controller.cpp
```diff
@@ -26,19 +29,204 @@ void PlatformerControllerSystem::SpinUp()
 
 void PlatformerControllerSystem::Run()
 {
-    Engine::Registry().view<InputReceiver, Sprite, Animator, PlatformerCharacter>().each(
-        [](const InputReceiver input_, Sprite& sprite_, Animator& animator_, const PlatformerCharacter& char_)
+    Engine::Registry().view<InputReceiver, Sprite,Transform, Animator, PlatformerCharacter>().each(
+        [](const InputReceiver input_, Sprite& sprite_, Transform& transform_,Animator& animator_,PlatformerCharacter& char_)
```
DPrivalov, line 33: might be useful to write smaller functions and call them here. But wait for today's lecture, we will discuss similar thing today

source/dagger/gameplay/platformer/platformer_main.h
```diff
@@ -1,12 +1,89 @@
 #pragma once
 
-#include "core/core.h"
+
 #include "core/game.h"
+#include "core/core.h"
```
DPrivalov, line 4: you need to try to use as less includes as possible in .h files.

This thing might be used here:
https://en.wikipedia.org/wiki/Forward_declaration
u are basically adding a prototypes and add includes with implementations in cpp

DPrivalov (changes requested):

source/dagger/gameplay/platformer/platformer_main.h
Coakim: I moved the definition of the methods to the .cpp file , but if I use the forward declaration I can't use the Character struct methods within other .cpp files. So I had to leave some includes in the .h file , although I'm sure there is a better way to do this  , any suggestions?

source/dagger/gameplay/platformer/platformer_combat.cpp
Coakim: You're right! Added checking if character has any animation active

source/dagger/gameplay/platformer/platformer_main.h
Coakim: So , I found an article about this here umich.edu/~eecs381/handouts/IncompleteDeclarations.pdf .
What I did was use forward declaration of struct in platformer_main.h , but I had to put the full definition of it in character_complete_definition.h , which still has the problem of it being a .h file with multiple #includes in it , but it is only used in .cpp files which need to acces the methods of the Character struct and not the other .h files . Is that any better?

Coakim: Actually , I had no idea why I tried getting the Character struct inside of platformer_combat.h ... I didn't need it at all , I  already had the CharacterHealth component that I needed. I fixed the code and returned the includes the way they were before. Sorry about that , I guess it served as a example in todays lecture :D

DPrivalov (approved):
now it's nice! 
good example for other teams as well.

ask your teammates to review it before merging

UbiBelETF:
> Added more movement animations to the platformer game , as well as some attack animations (jumping , rolling , heavy , light , down attack). Players can use jump while in the air to perform a double jump. Rolling is performed by tapping the same key used for running 2 times fast, which performs a roll in the selected direction.Implemented a simple combat system which tracks both players health points (max 100) , and the health bars for both players are displayed on top of the screen. Used the existing collision system to detect collision , and if the collision is detected while an attack is performed the players on the other end of the attack takes damage based on the type of the attack performed. The damage is visibly displayed as the health bar drains.
> If the player reaches 0 health he dies , making that player unable to recieve any input (Didn't implement stopping the game when the player dies yet).
> *Was not able to make the health bars follow the camera properly (It is possible for health bars to dissapear from the view).Tried using the parallax system but failed ...

What we want to do with UI is to render outside of the camera, I'll be working on this by the end of the week. It'll be one more parameter in `Sprite`, so that you can basically say `sprite.layer = SpriteLayer::ScreenSpace` and that's it.

PR #70, opened 2020-11-21
Custom collision system, players can't go through each other.

senadkurtisi:
This custom collision system prevents the players from colliding into each other (currently not working for jumping on the objects), the system first checks if the collision happened and remembers all the sides from which the collision happened. Each collision component contains information which tells us if we should take the collision between two entities (collidesWith) based on the entityType attribute. Currently both players have entityType=0 for testing purposes. In the future distinction between the classes (player, enemy, collectable) which is remebered in the entityType will be most likely controlled by enum

### DIFF
--- a/source/dagger/gameplay/platformer/platformer_collision.h
+++ b/source/dagger/gameplay/platformer/platformer_collision.h
@@ -23,7 +23,7 @@ struct PlatformerCollision
     
     // player, terrain, enemies, traps, collectables, weapons
     // For player it would be: false, true, true, true, ???, false
-    std::array<Bool, 6> collidesWith { true, true, true, true, true, false };
+    StaticArray<Bool, 6> collidesWith { true, true, true, true, true, false };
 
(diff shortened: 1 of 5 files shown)


source/dagger/gameplay/platformer/platformer_collision.h
```diff
@@ -0,0 +1,44 @@
...
+#include "gameplay/platformer/platformer_controller.h"
+
+using namespace dagger;
+using namespace platformer;
+
+struct CollisionInfo
+{
+    Bool hasCollided{ false };
+    // 0: right | 1: left | 2: top  |  3: down
+    std::array<Bool, 4> collisionSide{ false, false,false,false };
```
UbiBelETF, line 15: `StaticArray<Bool, 4>` is the Dagger way to go :) if we replace this with something in the future, it'll be easier to use the abstracted version.

UbiBelETF (approved):

UrosBorovcanin (approved):
The code is written well and functions as intended.

PR #73, opened 2020-11-22
Engine/sprite atlas

UbiBelETF:
Added support for tiled spritesheets. The format for the spritesheet is a simple custom textual format and not json, for simplicity. The format can contain the following lines:

- <blank> (skipped in parsing)
- NAME X Y W H (builds a frame called NAME, starting at (X, Y) with dimensions (W, H)
- NAME X Y W H A (builds A frames called NAME:N, where N is between 1 and A, from (X + W * (N - 1), Y) with dimensions (W, H), used for creating multiple frames of the same dimensions, in the same row, under the "same" name)

The texture names generated from this file always start with "spritesheets:SPRITESHEET_NAME:NAME".

The system for parsing spritesheets is done in `sprite_render.cpp`, but adding this system requires a lot of changes elsewhere.


source/dagger/core/graphics/sprite.h
```diff
@@ -21,8 +37,16 @@ namespace dagger
 		ViewPtr<Shader> shader{ Shader::s_FirstLoadedShader };
 	};
 
-	void AssignSpriteTexture(Sprite& spriteTarget_, String textureName_);
-	void AssignSpriteTexture(Sprite& spriteTarget_, ViewPtr<Texture> texture_);
+	struct Spritesheet
```
DPrivalov: I don't have a proper name for this, but seems like Spritesheet is a bit misleading

source/dagger/main.cpp
```diff
@@ -1,14 +1,16 @@
...
 #include "core/engine.h"
-#include "gameplay/ping_pong/ping_pong_main.h"
-#include "gameplay/platformer/platformer_main.h"
-#include "gameplay/racing/racing_main.h"
-#include "gameplay/team_game/team_game_main.h"
+//#include "gameplay/ping_pong/ping_pong_main.h"
+//#include "gameplay/platformer/platformer_main.h"
+//#include "gameplay/racing/racing_main.h"
+//#include "gameplay/team_game/team_game_main.h"
+#include "gameplay/tiles_example/tiles_example_main.h"
```
DPrivalov, line 7: why do u need to comment everything else here? 
it'll make enabling of other examples a bit harder

DPrivalov (approved):

SofijaBosk (approved):

milosmilosevic13 (approved):

source/dagger/core/graphics/sprite.h
UbiBelETF: Agreed! Let's think about this collectively :D

source/dagger/main.cpp
UbiBelETF: I just wanted to get the game to load faster :D we can get this out, really

source/dagger/core/graphics/sprite.h
UbiBelETF: SpriteFrame!

PR #74, opened 2020-11-22
Created Character Controller

Yugo0:
I implemented finite state machine using ECS framework.  I'm wondering if there is a way to improve this, or is it good enough?


data/input-contexts/among_them.json
```diff
@@ -0,0 +1,15 @@
+{
+	"context-name": "AmongThem",
```
DPrivalov: might be better to call these:
AmongThem_input
among_them_input.json

source/dagger/gameplay/team_game/character_controller_states/idle_state.cpp
```diff
@@ -0,0 +1,43 @@
+#include "idle_state.h"
+
+#include "core/engine.h"
+#include "core/input/inputs.h"
+
+#include "gameplay/team_game/character_controller.h"
+
+void IdleStateSystem::OnInitialize(Registry& registry_, Entity entity_)
+{
+	InputReceiver& receiver = registry_.get<InputReceiver>(entity_);
```
DPrivalov: @gavrilovmiroslav , take a look. We are doing this in 2 systems, maybe we should move it somewhere?

DPrivalov (approved):
Nice and clean! good job.

I rly like the size of this pull request, there is no any non-related changes here.

Yugo0:
Concerning the `OnInitialize` part, I tried to put it in `CharacterControllerSystem` that existed, but it didn't work, so I just decided to go with this.

source/dagger/gameplay/team_game/character_controller_states/idle_state.cpp
UbiBelETF: I'd actually propose something a bit different, if there's interest. Instead of making engine-level systems for states, I'd go for creating a `FiniteStateMachine` class that will basically do the same thing that the engine does (so, have a list of systems and cycle through them, but with the exception of only cycling through the active ones). Then both of these systems could be translated to a `State` base class that does the same thing, but isn't on the same level as the actual systems. Instead, the base class (`FiniteStateMachine`) could be that and hold all of that state within.

What I'm afraid of is polluting the systems loop with basically idling states from different state machines, instead of having just a `PlayerBehaviourSystem` which, internally, handles a sub-engine, and then an `AIBehaviourSystem`, etc,

What do you think?

kostaDimitrijevic (approved):

UbiBelETF:
I just have to say... what you implemented isn't what I said :D

UbiBelETF:
So, I've gone in and done some editing, just to show you how I'd have done it. So, the FSM was supposed to be an abstraction mechanism you can reuse, not just copy-pasted code from `System`. Look at `character_controller` class for example of usage. It's much shorter and less boiler-platey, although the core itself is admittedly quite complex.

Yugo0 (commented):
Why do we keep `fsm.cpp` file?

Yugo0 (commented):
Looks good.

source/dagger/gameplay/team_game/character_controller.cpp
```diff
@@ -0,0 +1,72 @@
...
+}
+
+// ----------------------------------------------------------
+// fsm stuff
+// idle
+
+void Idle::Run(CharacterController& ctrl_)
+{
+	// it's unsafe to compare floats precisely so we use epsilon
+	if (glm::epsilonNotEqual(glm::length(ctrl_.userInput), 0.0f, CHAR_CONTROLLER_EPSILON))
```
DPrivalov: mb we can move it to some dagger file: glm::epsilonNotEqual

DPrivalov (approved):
this fsm thing looks a bit complicated, but seems fine

source/dagger/gameplay/team_game/character_controller.cpp
gavrilovmiroslav: It's always harder to abstract away functions especially glm ones, because of their complex signatures. Maybe we could do macros for simple stuff like this

gavrilovmiroslav:
> this fsm thing looks a bit complicated, but seems fine

it is, I'm spending some time to mask it a bit, as there are some parts that don't need to be so exposed.

kostaDimitrijevic (approved):

source/dagger/gameplay/team_game/character_controller.h
```diff
@@ -0,0 +1,67 @@
+#pragma once
+
+#include "core/system.h"
+#include "gameplay/common/fsm.h"
+#include "core/game/transforms.h"
+
+#include <utility>
+
+using namespace dagger;
+
```
DPrivalov, line 10: we need game namespace here

DPrivalov (commented):
one thing and it's good

PR #75, opened 2020-11-22
Added text rendering to engine. Added TextKnife, the font spritesheet prep tool.

UbiBelETF:
Uses sprite atlases extensively, so merging into that branch first.


DPrivalov:
we should change base to main, before merging.

source/dagger/core/graphics/text.cpp
```diff
@@ -0,0 +1,51 @@
+#include "text.h"
+
+#include "core/core.h"
+#include "core/engine.h"
+#include "core/graphics/sprite.h"
+
+using namespace dagger;
+
+void Text::Set(String font, String message_, Vector3 pos_)
```
DPrivalov, line 9: font_ =)

source/dagger/core/graphics/text.cpp
```diff
@@ -0,0 +1,51 @@
...
+{
+	this->font = font;
+
+	auto& registry = Engine::Registry();
+
+	assert(Engine::Res<Texture>().contains(fmt::format("spritesheets:{}", font)));
+
+	if (entities.size() > 0)
```
DPrivalov, line 17: where do u set these?

DPrivalov: we might need to do this removal differently (or add some warning if text is set in the middle of the frame)

DPrivalov (approved):

source/dagger/core/graphics/text.cpp
UbiBelETF: i thought we were doing a helper for this adding and removing, so decided to just do it the old way and then move it once we establish how we're going to work it in. should i change it right now?

UbiBelETF: also, i set them up in the `Set` function!

UbiBelETF:
> we should change base to main, before merging.

maybe we should merge this into sprite atlas first, and then merge sprite atlas into main; that was my idea. these changes are smaller than the ones i have to do there, and doing them there and then implementing here might be harder.

source/dagger/core/graphics/text.cpp
UbiBelETF: Also, I found this: https://github.com/skypjack/entt/issues/98

It seems that we are somehow hitting some behaviour which isn't really dangerous?

PR #77, opened 2020-11-23
Features for team game, small enhancement for collisions

milosmilosevic13:


mmacura9 (approved):

source/dagger/gameplay/team_game/team_game_main.cpp
```diff
@@ -53,3 +56,48 @@ void team_game::SetupWorld(Engine &engine_)
         col.size = sprite.size;
     }
 }
+
+struct Mage {
```
DPrivalov: strange that Mage struct is inside main, and not inside mage.h

source/dagger/gameplay/common/simple_collisions.cpp
```diff
@@ -30,9 +30,11 @@ void SimpleCollisionsSystem::Run()
             {
                 collision.colided = true;
                 collision.colidedWith = *it2;
+                collision.colisions.emplace_back(*it2);
```
DPrivalov, line 43: don't forget to clean this list at the beginning of this function. You can reset colidedWith as well.

just add one more loop before 'while(it != view.end())'

DPrivalov (approved):
simple collision changes should be a separate pull request, but it's ok this time.

source/dagger/gameplay/team_game/team_game_main.cpp
milosmilosevic13: It was there initially just for some testing, it'll be moved.

milosmilosevic13:
Needed to test mage attacking and range of attack, so I had to add some things that should have been a separate features/pull requests. Last commit of this pull.

nkristina (approved):

PR #78, opened 2020-11-23
Added a foundation for a combat system for further development

Coakim:
Implemented the foundation of a combat system for the game. Currently , the combat system tracks health and stamina and visually displays the changes in both. (Currently health drops on collision between characters and stamina drains from running , for example purposes , will be changed)

Will further develop and modify as the player controller system gets implemented with the state dessign pattern. Moved simple collision system to plight_collisions system , changed the colidedWith entity to a list of entities in case of multiple collisions and added the loop that resets all collisions before the next check of collisions.

### DIFF
--- a/source/dagger/gameplay/plight/plight.cpp
+++ b/source/dagger/gameplay/plight/plight.cpp
@@ -95,7 +95,7 @@ void Plight::WorldSetup(Engine &engine_)
     camera->position = { 0, 0, 0 };
     camera->Update();
 
-    plight::SetupWorld_test1(engine_);
+    plight::SetupWorld_CombatSystem(engine_);
 }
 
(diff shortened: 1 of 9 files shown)


source/dagger/gameplay/plight/plight.cpp
```diff
@@ -26,6 +29,9 @@ struct PlightCharacter
     Animator& animator;
     InputReceiver& input;
     TopdownCharacter& character;
+    PlightCollision& col;
+    Transform& transform;
+    CombatStats& cstats;
 
     static PlightCharacter Get(Entity entity)
```
UbiBelETF: Styleguide: parameters should always have a tracing underscore, eg. `Entity entity_`

UbiBelETF (approved):
Seems cool!

PR #79, opened 2020-11-25
Implemented level generation and transition using .txt files

senadkurtisi:
GameManagerSystem loads the according .txt file(based on current level) and reads which texture it needs to load (and from which directory) along with some other parameters which are necessary for scaling and positioning. 
I implemented a simple Test player input system which allowed me to test if the transitioning between levels is possible. On the press of key P the levels should change. Currently only three simple levels are implemented.

Levels are simple considering the fact that not everything about the styles of enemies is fully developed and level generation in this form should be used as a base for implementing the custom collision and character controller. 
Level generation will be improved progressively.


UbiBelETF (commented):
I'd suggest moving to working on the editor, as it seems like you're really missing that functionality! However, if you think this does the job, and are happy with this process, it _is_ your game.

source/dagger/gameplay/team_game/team_game_main.cpp
```diff
@@ -1,4 +1,5 @@
 #include "team_game_main.h"
+#include <fstream>
```
DPrivalov, line 3: do u need this here?

source/dagger/gameplay/team_game/game_manager.cpp
```diff
@@ -0,0 +1,99 @@
...
+void GameManagerSystem::OnEndOfFrame()
+{
+    auto levelComponentsView = Engine::Registry().view<ComponentOfLevel>();
+
+    for (auto& entity : levelComponentsView)
+    {
+        auto& levelComponent = levelComponentsView.get<ComponentOfLevel>(entity);
+        if (levelComponent.discard)
+        {
+            Engine::Registry().remove_all(entity);
```
DPrivalov: my suggestion is to do complete removal here, and load another level after that.
(check racing reset for example)

that way it'll be easier to clean everything properly and speed up the removal process. 

You might need to save some player states or something between levels. But I think you will be able to do that even with full entities reset.

++ @gavrilovmiroslav

DPrivalov (approved):
might be ok as a first version.

source/dagger/gameplay/team_game/team_game_main.cpp
senadkurtisi: Oh, actually no, I don't. Loading from .txt was first tested for a single level in the team game main, and later modified to achieve transition between levels and therefore the code was moved to a different .cpp, but I forgot to remove that include. Thanks :)

UrosBorovcanin (approved):
The changes run well. They are good for now, but in the future we might need more modifiers in the .txt file to change certain component fields. Like making platforms that can move or making enemies with specific properties

senadkurtisi:
> The changes run well. They are good for now, but in the future we might need more modifiers in the .txt file to change certain component fields. Like making platforms that can move or making enemies with specific properties

Good point! This is the initial version so it is not so robust, but in the future we should have more distinct functions which should handle the loading of the enemies and collectables for example separately. Currently collectables are being loaded as a background. 

Also, the `LoadTextures` function would need to have more arguments, in order to be able to add components to the entities like a _TrapComponent_ etc.

FedjaVujicic (approved):
Nicely designed, and works well

PR #82, opened 2020-11-25
Team/morning afternoon/feature/aiming system

Coakim:
Made a simple aiming system along with circular movement based on it as a possible solution to how we are going to handle combat in our game. Aiming system has a crosshair component which rotates around the player character sprite (angle changes on A and D keys) , and the player now only has the ability to go forward or backwards , but in the direction the crosshair is pointing at , so basically 360 degrees. (Player sprite scale also changes based on the viewing angle , so it kind off looks nice even though the game is sort off three-quarter view)
This might make it easier for 2 players to play on the same keyboard (as it is currently the plan ) , because movement and aiming would be done by one hand , leaving other hand free for using spells and other combat related things


source/dagger/gameplay/plight/plight.cpp
```diff
@@ -80,7 +91,8 @@ void Plight::GameplaySystemsSetup(Engine &engine_)
 {
     engine_.AddSystem<TopdownControllerSystem>();
     engine_.AddSystem<PlightCollisionsSystem>();
-    engine_.AddSystem<PlightCombatSystem>();
+    //engine_.AddSystem<PlightCombatSystem>();
```
DPrivalov: are u sure u need to comment it out?

source/dagger/gameplay/plight/topdown_controller.cpp
```diff
@@ -29,10 +36,9 @@ void TopdownControllerSystem::SpinUp()
 
 void TopdownControllerSystem::Run()
 {
-    Engine::Registry().view<InputReceiver, Sprite,Transform, Animator, TopdownCharacter, CombatStats>().each(
-        [](const InputReceiver input_, Sprite& sprite_,Transform& transform_, Animator& animator_,TopdownCharacter& char_, CombatStats& cstats_)
+    Engine::Registry().view<InputReceiver, Sprite,Transform, Animator, TopdownCharacter, CombatStats , PlightCrosshair>().each(
+        [](const InputReceiver input_, Sprite& sprite_,Transform& transform_, Animator& animator_,TopdownCharacter& char_, CombatStats& cstats_, PlightCrosshair& crosshair_)
         {
-            Float32 moveX = input_.values.at("moveX");
             Float32 moveY = input_.values.at("moveY");
```
DPrivalov: might be better to call it moveForward everywhere

DPrivalov (approved):
Looks good

source/dagger/gameplay/plight/plight.cpp
Coakim: No need , I just commented it out so that I can test the new movement and aiming without draining stamina while running. It works fine with the combat system as well , I just forgot to remove the comment , will do !

source/dagger/gameplay/plight/topdown_controller.cpp
Coakim: True , it doesn't really make sense to call it moveY anymore . Will change it :)

Coakim: Changed it to just move , since we can also go backwards :D

veljko-dj (approved):
I tried this aiming system. It looks good to me. And that's great for multiplayer on the same computer until the controllers are added

DamjanTM (approved):
Everything seems to be working

PR #85, opened 2020-11-26
Added spritesheets and animations

Yugo0:
Created .spritesheet files in data/spritesheets folder and .json files in data/animations/among_them folder. Also, tested idle and running animation by adding them to the player character, so he is animated now.


source/dagger/gameplay/team_game/character_controller.cpp
```diff
@@ -55,6 +60,17 @@ void Running::Run(CharacterController& ctrl_)
 		return;
 	}
 
+	// Character orientation
+	if (ctrl_.userInput.x > 0.0f)
+	{
+		ctrl_.sprite->scale.x = 3;
```
DPrivalov, line 66: looks like 3 is magic number here. It's better to use multiply by -1 here

DPrivalov (approved):

source/dagger/gameplay/team_game/character_controller.cpp
Yugo0: I tought about that, but if I did that, then i would have to check for the value of userInput, and then to check the current value of sprite to see whether it needs to be multiplied by -1. I'm not sure, but it feels like making code needlessly complicated.
I put this in here just to test funtionality of animations, that 3 will probably go into some variable (i. e. characterSize) in the future.

smiljanaspasic (approved):

PR #87, opened 2020-11-27
Added shooting, bullets follow mouse cursor

TeodorVasiljevic365:


source/dagger/gameplay/team_game/team_game_main.cpp
```diff
@@ -91,7 +93,7 @@ void lab::SetupWorld(Engine &engine_)
     auto& reg = engine_.Registry();
 
     float zPos = 1.f;
-
+    /*
```
DPrivalov, line 96: it's better to get rid from that code completely, no need to comment it out.

DPrivalov (approved):

SofijaBosk (approved):

PR #89, opened 2020-11-27
Team/delta snakes/feature/character controller

FedjaVujicic:
Created a simple character controller system for the team game using FSM


DPrivalov (changes requested):
wait for Miroslav to finish fsm in main, it'll be better not to copy-paste it twice

https://github.com/UbiBelETF/dagger/pull/92

source/dagger/gameplay/team_game/controller_fsm.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+	Float32 run = input.Get("run");
+
+	if (EPSILON_ZERO(run))
+	{
+		GoTo(ECharacterStates::Idle, state_);
+	}
+	else
+	{
+		sprite.scale.x = run;
```
senadkurtisi, line 56: Considering the fact that the sprites we have need to be scaled (scale would be definitely tuned) in order to fit nicely into the ambient of the level scene, we should probably in the future change this with respect to some previously stored value for the scale of the main character. We should note this somewhere in case we forget. 😄

senadkurtisi (approved):
Looks good to me! Good to merge!

UrosBorovcanin (approved):
The FSM works well and seems simple enough to work with.
The code is clear and i dont see any irregularities.

PR #91, opened 2020-11-27
Implemented controller system in a finite state machine pattern

Coakim:
Finite state machine implementation of the character controller system (ECS style - every state is presented by a separate system). Currently has only run and idle states , but we plan on upgrading it further.(Uses the old movement system until the circular movement in other pull request is approved by teammates, will adapt it to that if it is approved).


DPrivalov (approved):
looks fine for me, let's see how it'll look with more states

++ @gavrilovmiroslav

PR #92, opened 2020-11-28
Created full FSM implementation and implemented a platformer example …

UbiBelETF:
…using this system.

To get the gist of how to use this system, look at:
 
 - gameplay/platformer/platformer_controller.cpp -- to see the new implementation
 - gameplay/platformer/character_controller_fsm.h -- to see how a FSM is defined
 - gameplay/platformer/character_controller_fsm.cpp -- to see how a FSM is actually implemented

For a deep dive, look at:

- core/game/finite_state_machines.h -- all the magic happens here


source/dagger/gameplay/platformer/character_controller_fsm.cpp
```diff
@@ -0,0 +1,61 @@
...
+// Idle
+
+
+void CharacterControllerFSM::Idle::Enter(CharacterControllerFSM::StateComponent& state_)
+{
+	auto& animator = Engine::Registry().get<Animator>(state_.entity);
+	AnimatorPlay(animator, "souls_like_knight_character:IDLE");
+}
+
+DEFAULT_EXIT(CharacterControllerFSM, Idle);
```
DPrivalov, line 20: maybe it'll be better to add flags in DEFINE_STATE, to add default implementation.
not sure if this will be better thou, so it's just an idea

DPrivalov (approved):

source/dagger/gameplay/platformer/character_controller_fsm.cpp
gavrilovmiroslav: this way, if you want to add a non-default implementation, you change one file, not two, but othwerise is the same

PR #98, opened 2020-11-29
Added a physics system that has a gravity feature

Ranko030:


Ranko030:
Also added an ability that a hero can have a double jump.

source/dagger/gameplay/team_game/brawler_controller.cpp
```diff
@@ -27,88 +30,73 @@ void BrawlerControllerSystem::SpinUp()
...
-            if(!char_.jump || char_.attack){
-                if (char_.fall)
+           // Float32 attack = input_.values.at("light");
+
+            if (run != 0) {
+                sprite_.scale.x = run;
+                physics_.velocity.x += sprite_.scale.x* char_.speed;
+                char_.run = true;
+            }
+            /*if(!char_.jump )
```
UbiBelETF: Why keep such a large comment? It's confusing to review, this way.

Ranko030: Oh sorry forgot to delete that.

DPrivalov:
Check SimpleCollisions.cpp  file, SimpleCollisionsSystem::Run should reset collision.colided status before starting to process new frame.

Ranko030:
> Check SimpleCollisions.cpp file, SimpleCollisionsSystem::Run should reset collision.colided status before starting to process new frame.

Well, I thought of removing the SimpleCollisionsSystem and just using the SimpleCollision struct in CollisionSystem to see if there will be a collision in the next frame. Now if there will be a collision then just try to lower the velocity to the point when there won't be one.

DPrivalov:
@Ranko030 , cool
mention me in comments when you implement it.

Ranko030:
@DPrivalov check the collisions.cpp file. In there I implemented that system . That is the thing I added from the team/tired_bunch/test/gravity branch when I fixed it.

DPrivalov (approved):

PR #101, opened 2020-11-29
Engine/camera depth scale fix

### LINKED ISSUE #94: Far clipping plane distance reduces when camera zoom increases
When zoom = 1, sprites in scene stop being rendered at z = 100;
zoom = 2, at z = 51;
zoom = 3, at z = 34.

It seems that the far clipping plane distance gets divided by zoom value at some point.

UbiBelETF:
Fixes #94 
Please test.

### DIFF
--- a/source/dagger/core/input/inputs.cpp
+++ b/source/dagger/core/input/inputs.cpp
@@ -271,12 +271,13 @@ void InputSystem::Run()
 					}
 				}
 
-				for (auto& [key, value] : receiver_.values)
+				Map<String, Float32> newValues;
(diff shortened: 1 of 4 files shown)


source/dagger/core/input/inputs.cpp
```diff
@@ -271,12 +271,13 @@ void InputSystem::Run()
 					}
 				}
 
-				for (auto& [key, value] : receiver_.values)
```
DPrivalov, line 274: hm... strange that this one is here

DPrivalov (approved):

PR #102, opened 2020-11-29
bullets are no longer spawned at the center of the character

TeodorVasiljevic365:


SofijaBosk (approved):

source/dagger/gameplay/team_game/shoot.cpp
```diff
@@ -24,8 +24,9 @@ void lab::CreateBullet(Vector2 position, Float32 ratio, Vector2 directions)
 	AssignSprite(sprite, "Blob");
 	sprite.size = { 10, 10 };
 	auto& transform = reg.emplace<Transform>(entity);
-	transform.position.x = position.x;
-	transform.position.y = position.y;
+	Float32 distance = sqrt((11 * 11) / (1 + ratio * ratio));
+	transform.position.x = position.x + (distance * directions.x);
+	transform.position.y = position.y + (distance * ratio * directions.y * -1);
```
DPrivalov, line 29: if it's working then it's fine.

I suggest to simply add some variable with char size and add it here instead of distance.

DPrivalov (approved):

PR #105, opened 2020-11-29
Added simple aiming system

Coakim:
Simple aiming system

### DIFF
--- a/source/dagger/gameplay/common/aiming_system.cpp
+++ b/source/dagger/gameplay/common/aiming_system.cpp
@@ -24,6 +24,20 @@ void AimingSystem::Run()
         {
             Float32 rotate = input_.values.at("rotate");        //Get the input value of the rotate command - the amount the angle changes if the button is pressed 
 
+        /*    
+        *            Example setup of for "rotate" command in input-context
+                     "command-name": "rotate",
+                        "actions" : [
+                      {
+                        "trigger": "KeyA",
+                        "value" : 0.05
+                      },
+                      {
+                       "trigger": "KeyD",
+                       "value" : -0.05
+                      }
+                ]*/
+
             if (rotate) { 
                 crosshair_.angle += rotate;   //Change the angle of the crosshair
 
--- a/source/dagger/gameplay/common/aiming_system.cpp
+++ b/source/dagger/gameplay/common/aiming_system.cpp
@@ -22,21 +22,23 @@ void AimingSystem::Run()
     Engine::Registry().view<InputReceiver, Sprite, Crosshair>().each(
         [](const InputReceiver input_, Sprite& sprite_, Crosshair& crosshair_) //The sprite component with the same entity as this crosshair component is of a sprite that is used for center of rotation (like character for example)
         {
+        /*
+         Example setup of for "rotate" command in input-context :
+
+                    "command-name": "rotate",
+                       "actions" : [
+                     {
+                       "trigger": "KeyA",
+                       "value" : 0.05
+                     },
+                     {
+                      "trigger": "KeyD",
+                      "value" : -0.05
+                     }
+                    ]
+       */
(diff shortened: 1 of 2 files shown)


source/dagger/gameplay/common/aiming_system.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+#include "core/graphics/sprite.h"
+#include "core/input/inputs.h"
+
+#include <cmath>
+#include <math.h>
+
+void AimingSystem::OnInitialize(Registry& registry_, Entity entity_)
+{
+    InputReceiver& receiver = registry_.get<InputReceiver>(entity_);
```
DPrivalov, line 13: @gavrilovmiroslav , do we still need this?

source/dagger/gameplay/common/aiming_system.cpp
```diff
@@ -0,0 +1,70 @@
...
+        receiver.values[command] = 0;
+    }
+}
+
+void AimingSystem::Run()
+{
+    Engine::Registry().view<InputReceiver, Sprite, Crosshair>().each(
+        [](const InputReceiver input_, Sprite& sprite_, Crosshair& crosshair_) //The sprite component with the same entity as this crosshair component is of a sprite that is used for center of rotation (like character for example)
+        {
+            Float32 rotate = input_.values.at("rotate");        //Get the input value of the rotate command - the amount the angle changes if the button is pressed
```
DPrivalov, line 40: add a comment above with "rotate" setup from ini file, so other teams can easily set it up.

DPrivalov (approved):

source/dagger/gameplay/common/aiming_system.cpp
```diff
@@ -24,6 +24,20 @@ void AimingSystem::Run()
...
+                        "actions" : [
+                      {
+                        "trigger": "KeyA",
+                        "value" : 0.05
+                      },
+                      {
+                       "trigger": "KeyD",
+                       "value" : -0.05
+                      }
+                ]*/
```
DPrivalov: */ - move this to new line

source/dagger/gameplay/common/aiming_system.cpp
```diff
@@ -24,6 +24,20 @@ void AimingSystem::Run()
         {
             Float32 rotate = input_.values.at("rotate");        //Get the input value of the rotate command - the amount the angle changes if the button is pressed 
 
+        /*    
+        *            Example setup of for "rotate" command in input-context
```
DPrivalov: remove *, and move to the left a bit, like this:

Example setup of for "rotate" command in input-context:

DPrivalov (approved):
there is one minor thing and good to go

source/dagger/gameplay/common/aiming_system.cpp
DPrivalov: Also u can move this comment before this line:

 Float32 rotate = input_.values.at("rotate");

Coakim: Will do 👍

source/dagger/gameplay/common/aiming_system.cpp
UbiBelETF: No, this was changed recently, in the same event that broke Mac. We no longer need this, but the `Get` function needs to be used from `InputReceiver`. So: no setting commands to 0 manually here

source/dagger/gameplay/common/aiming_system.cpp
```diff
@@ -0,0 +1,86 @@
...
+                       "trigger": "KeyA",
+                       "value" : 0.05
+                     },
+                     {
+                      "trigger": "KeyD",
+                      "value" : -0.05
+                     }
+                    ]
+       */
+            Float32 rotate = input_.values.at("rotate");        //Get the input value of the rotate command - the amount the angle changes if the button is pressed
```
UbiBelETF, line 40: Use `input_.Get("rotate")` instead. Maybe the `values` map should be private in the future...

PR #107, opened 2020-11-30
Added camera tracking

Yugo0:


source/dagger/gameplay/team_game/camera.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+#include "core/system.h"
+
+using namespace dagger;
+
+class CameraSystem : public System
```
DPrivalov: don't forget to use team_game namespace here

DPrivalov (approved):

source/dagger/gameplay/team_game/camera.h
Yugo0: To put `CameraSystem` in `team_game` namespace?

DPrivalov: yup, every class/struct from gameplay/team_game

PR #108, opened 2020-11-30
Fix for Aiming System feature

Coakim:
Removed OnInitialize method , changed to inputs.h Get method instead of values.at .


DPrivalov (approved):
@gavrilovmiroslav , this one is for u =)

PR #111, opened 2020-11-30
Engine/fix/world coordinates

UbiBelETF:
Test this out, and also take a look in the platformer example: it now spawns the goblins on click!

Also check if I broke something else in the process.


UbiBelETF:
Also, please please *don't* merge this. I pulled your team version, but will isolate the last commit for engine later. Just wanted to not cause you any problems by checking this out. :)

PR #119, opened 2020-12-03
Team/morning afternoon/feature/tile maps

Coakim:
Integrated tilemaps into the game , and made a simple example arena for further systems testing.

*Also created a fixed update interval for aiming system (FPS now doesn't affect speed of crosshair rotation)


DPrivalov (approved):

DamjanTM (commented):
The functions in plight_tilemaps_initialization.cpp should probably be turned into a factory.
Though I'm not sure how to do this with the more specific tiles

PR #122, opened 2020-12-03
Aiming System

Coakim:
FPS now doesn't affect speed of crosshair rotation

### DIFF
--- a/source/dagger/gameplay/common/aiming_system.cpp
+++ b/source/dagger/gameplay/common/aiming_system.cpp
@@ -15,54 +15,57 @@ void AimingSystem::Run()
     Engine::Registry().view<InputReceiver, Sprite, Crosshair>().each(
         [](InputReceiver input_, Sprite& sprite_, Crosshair& crosshair_) //The sprite component with the same entity as this crosshair component is of a sprite that is used for center of rotation (like character for example)
         {
-        /*
-         Example setup of for "rotate" command in input-context :
+            crosshair_.currentTimer += Engine::DeltaTime();
+            if (crosshair_.currentTimer >= crosshair_.rotationTimer) {
+                /*
+                 Example setup of for "rotate" command in input-context :
 
-                    "command-name": "rotate",
-                       "actions" : [
-                     {
-                       "trigger": "KeyA",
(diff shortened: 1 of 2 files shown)


DPrivalov:
could you please explain your idea here? I don't understand why we cannot have rotationSpeed

Coakim:
> could you please explain your idea here? I don't understand why we cannot have rotationSpeed

Well , currently the rotation input value acts as a rotation speed (it's value is added to the crosshair angle value every iteration of the loop) , but I had situations where I was getting around 150fps and the crosshair movement was smooth , and then situation when i was getting 300+ fps and the crosshair rotation speed was way too fast.I figured out that is due to the execution speed of the program itself (how fast the game loop runs) , so basically I found out that the solution would be to actually not update the crosshair position every iteration of the loop , but in a fixed time interval. In case of lower fps the update would happen every time because the time between two iterations would be greater than the updateTimer value I defined , but in case of higher speed of program execution the update wouldn't happen always but after certain number of iterations which makes crosshair rotation speed similar regardles of the fps . Hope this makes sense

DPrivalov:
Here is an easier implementation of splitting FPS with characters speed:
void RacingPlayerInputSystem::Run()
...
t.position.x += ctrl.input.x * car.horzSpeed * Engine::DeltaTime();

horzSpeed is some speed you need to define and by multiplying it with Engine::DeltaTime() you are making changes independent from PCs FPS.

You can check this for more details:
https://gameprogrammingpatterns.com/game-loop.html

Coakim:
> Here is an easier implementation of splitting FPS with characters speed:
> void RacingPlayerInputSystem::Run()
> ...
> t.position.x += ctrl.input.x * car.horzSpeed * Engine::DeltaTime();
> 
> horzSpeed is some speed you need to define and by multiplying it with Engine::DeltaTime() you are making changes independent from PCs FPS.
> 
> You can check this for more details:
> https://gameprogrammingpatterns.com/game-loop.html

Oh , I actually just now realize that this is used in the character controller system when moving the character and it solves the same issue . I didn't understand at first this is what you were suggesting. I changed the solution to this one !

PR #123, opened 2020-12-03
team/morning_afternoon/feature/CombatSystemUpgrade

Coakim:
Health and stamina bars now follows player and added a fixed update timer for health and stamina draining

### DIFF
--- a/source/dagger/core/graphics/window.cpp
+++ b/source/dagger/core/graphics/window.cpp
@@ -218,6 +218,8 @@ void WindowSystem::Run()
 	auto* window = m_Config.window;
 	Engine::Dispatcher().trigger<PreRender>();
 
+	glfwGetFramebufferSize(m_Config.window, &m_Config.windowWidth, &m_Config.windowHeight);
+
 	glViewport(0, 0, m_Config.windowWidth, m_Config.windowHeight);
 	glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
 	glClearColor(0.39f, 0.58f, 0.92f, 1.0f);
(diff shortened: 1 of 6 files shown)


DPrivalov:
Just to check, are u moving the character with transform changes? or by changing sprite position?

Coakim:
> Just to check, are u moving the character with transform changes? or by changing sprite position?

When moving the character I was chaning both transform and sprite positions , and the stat bars just by changing the sprite positions

DPrivalov:
Transform is updating sprite positions for the same entity. So no need to change character sprite position if you are using transform for this entity

Coakim:
> Transform is updating sprite positions for the same entity. So no need to change character sprite position if you are using transform for this entity

Oh , didn't know that ... thanks for letting me know , will remove the sprite position updates then

Coakim:
> Just to check, are u moving the character with transform changes? or by changing sprite position?

Okay , I have a question regarding the sprite and transform positions. Is it better to use transform component for chaning the position of sprites in general?  Currently I only used transform component with character position changes , and I just changed sprite positions for my crosshair sprites and health and stamina bar sprites

DPrivalov:
> > Just to check, are u moving the character with transform changes? or by changing sprite position?
> 
> Okay , I have a question regarding the sprite and transform positions. Is it better to use transform component for chaning the position of sprites in general? Currently I only used transform component with character position changes , and I just changed sprite positions for my crosshair sprites and health and stamina bar sprites

Let's discuss it during today's call. (question can be: when should we use transform and when we can change sprite position instead)
In short it's really dependent on your game design. I think other sprites can get their position from character and have some offsets, so there is no need to have multiple transform components.

DamjanTM (approved):
looks good

PR #124, opened 2020-12-04
Small enhancements

milosmilosevic13:
Added random offset to characters so they don't follow exact same route every time; improved depth perception by placing bottom most characters closer to the camera; added background


DPrivalov (approved):
looks fine

PR #125, opened 2020-12-04
Circular Collisions

milosmilosevic13:
SimpleCollision component now contains list of all entities it collided with. Also SimpleCollision now has options to pick between Rectangular and Circular hitbox. For Circular hitboxes size.x is now used as a diameter.


source/dagger/gameplay/common/simple_collisions.h
```diff
@@ -2,15 +2,26 @@
...
 
+enum class EHitbox {
+    Rectangular = 1,
+    Circular = 2,
+};
+
 struct SimpleCollision
 {
-    Vector2 size;
+    Vector2 size; // If the hitbox is rectangular, size represents height and width, if circular, diameter and angle of coverage (currently unused)
```
DPrivalov, line 15: you can use c++ union here maybe.

DPrivalov: it'll make code more readable

milosmilosevic13: It crossed my mind at some point that a union could be used, but I just wanted to make code functional at first.

PR #127, opened 2020-12-04
Added tile system and test level.

Nikolar97:
Added a tile system and a test level for it.


source/dagger/gameplay/team_game/hero_character.cpp
```diff
@@ -0,0 +1,6 @@
+#include "hero_character.h"
+
+hero_character::hero_character()
```
UbiBelETF: This...
a) isn't needed because the class doesn't exist
b) doesn't compile because of the class not existing

source/dagger/gameplay/team_game/tilemap_legends.cpp
```diff
@@ -0,0 +1,121 @@
...
+	return entity;
+}
+Entity CreateCrate(Registry& reg_, UInt32 x_, UInt32 y_, char type) {
+	Entity entity = reg_.create();
+	auto& sprite = reg_.emplace<Sprite>(entity);
+	sprite.scale = { scale,scale };
+	sprite.position = { x_ * x_step, y_ * y_step, 90 };
+	AssignSprite(sprite, "spritesheets:tiles_dungeon:crate");
+	return entity;
+}
```
UbiBelETF, line 159: There are many problems here but I want team members to review. Do not merge this as-is.

UbiBelETF (changes requested):

Becaan (approved):

PR #128, opened 2020-12-04
Team/fried cucumber/feature/character collider

rastkomiocinovic:


Becaan (dismissed):

source/dagger/gameplay/team_game/team_game_main.cpp
```diff
@@ -65,6 +73,11 @@ struct MainCharacter
 
         chr.character.speed = 100;
 
+        //Collision setup:        
+        chr.col.size = { 25, 28 };
+        chr.tm.position = { position_ , 0.0f };
+        CharacterCollisionSystem::MainCharacterEntity = entity;
```
DPrivalov: Just a suggestion:
instead of doing this you can create a MainCharacter component (empty struct) and then emplace it here. And in any place you need you can access it by 
auto view = Engine::Registry().view<MainCharacter>();

DPrivalov (dismissed):

PR #129, opened 2020-12-06
Team/sultans of swing/feature/shapeshifting system

smiljanaspasic:
**Instead of creating a new state for shape-shifting I decided to implement this by adding the shape-shifting ability to current states in character controller where the animator would act accordingly to the button that has been pressed.**

It also has a simple level of mine which we haven't merged to main yet


source/dagger/gameplay/team_game/character_controller.cpp
```diff
@@ -46,8 +55,11 @@ void Idle::Run(CharacterController& ctrl_)
 	{
 		GoTo(ECharacterState::Running, ctrl_);
 	}
-
-	AnimatorPlay(*ctrl_.animator, "among_them_animations:knight_idle");
+	if (ctrl_.shapeshiftInput.x == 1) { idle = "among_them_animations:goblin_idle"; running = "among_them_animations:goblin_run"; 	}
+	if (ctrl_.shapeshiftInput.y == 1) { idle = "among_them_animations:slime_idle"; running = "among_them_animations:slime_run";}
+	if (ctrl_.shapeshiftInput.z == 1) { idle = "among_them_animations:bat"; running = "among_them_animations:bat"; }
+	if (ctrl_.shapeshiftInput.w == 1) { idle = "among_them_animations:knight_idle"; running = "among_them_animations:knight_run"; }
```
Yugo0: Maybe you can move this to `CharacterControllerSystem::Run`, that way you only write this code once, and you don't need `shapeshiftInput`.
Also, enemies will have to know which shape character has is in to decide whether to start the chase, so that information should be stored somewhere and assigned here.
I really like this solution, though.

source/dagger/gameplay/team_game/character_controller.cpp
```diff
@@ -63,14 +75,17 @@ void Running::Run(CharacterController& ctrl_)
...
 	{
-		ctrl_.sprite->scale.x = -3;
+		ctrl_.sprite->scale.x = -1;
 	}
-
-	AnimatorPlay(*ctrl_.animator, "among_them_animations:knight_run");
+	if (ctrl_.shapeshiftInput.x == 1) { running = "among_them_animations:goblin_run"; idle = "among_them_animations:goblin_idle"; }
+	if (ctrl_.shapeshiftInput.y == 1) { running = "among_them_animations:slime_run"; idle = "among_them_animations:slime_idle"; }
+	if (ctrl_.shapeshiftInput.z == 1) { running = "among_them_animations:bat"; idle = "among_them_animations:bat"; }
+	if (ctrl_.shapeshiftInput.w == 1) { running = "among_them_animations:knight_run"; idle = "among_them_animations:knight_idle"; }
```
Yugo0: Same as above.

Yugo0 (changes requested):
Good job!

smiljanaspasic:
@Yugo0 
Ok,so I reworked this as you suggested.I put everything in CharacterControllerSystem::Run . But I also needed to put the Animator from the Running::Enter and Idle::Enter into Running::Run and Idle::Run because I needed to ensure smooth transition into different characters.So now we have empty  Running::Enter and Idle::Enter methods.I could not delete them completely and could not figure out why so please help
Also what are your suggestions about storing the information about current character shape if you have any?Where could I put that?

Yugo0:
Looks good.

DPrivalov:
@Yugo0 , u need to approve if it's fine.
@kostaDimitrijevic , take a look at it as well

kostaDimitrijevic (approved):

Yugo0:
@DPrivalov We had a little chat on Discord, there are some small changes that still need to be made.

smiljanaspasic:
@Yugo0 @kostaDimitrijevic 
Ok guys everything is ready for the final review.Take a look and let me know if I can merge it now

Yugo0 (approved):

PR #130, opened 2020-12-06
Team/fried cucumber/feature/character control

Becaan:
Added hero attack animation

### DIFF
--- a/source/dagger/gameplay/team_game/hero_controller_fsm.h
+++ b/source/dagger/gameplay/team_game/hero_controller_fsm.h
@@ -5,20 +5,30 @@
 enum struct EHeroStates
 {
 	Idle,
-	Running
+	Running,
+	Attacking
+};
+
+enum FacingPostion
+{
+	up,
+	down,
+	side
 };
 
 struct HeroControllerFSM : public FSM<EHeroStates>
 {
 	DEFINE_STATE(HeroControllerFSM, EHeroStates, Idle);
 	DEFINE_STATE(HeroControllerFSM, EHeroStates, Running);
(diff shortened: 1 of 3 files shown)


source/dagger/gameplay/team_game/hero_controller_fsm.h
```diff
@@ -5,20 +5,30 @@
...
 
 struct HeroControllerFSM : public FSM<EHeroStates>
 {
 	DEFINE_STATE(HeroControllerFSM, EHeroStates, Idle);
 	DEFINE_STATE(HeroControllerFSM, EHeroStates, Running);
+	DEFINE_STATE(HeroControllerFSM, EHeroStates, Attacking);
 
-
-
+	static bool stopAttackOnNextRepeat;
```
DPrivalov, line 25: could you please explain why these 2 should be static?

and explain the name as well please, it's not that clear to me what it should be used for

Becaan: stopAttackOnNextRepeat is a HeroControllerFSM's member and since we have only one hero making it static member shouldnt be a problem. Also, it is used in Run , Enter and Exit which are methods of State struct so it it easier accessing it without having to instantiate a HeroControllerFSM object.

stopAttackOnNextRepeat  does exactly what the name says, stops the attack animation on its next repeat. The idea behind it is that when u press the attack button u stay in "Attacking" state until the animation plays one full time because I had a problem that after pressing the attack button states change from attacking to idle too fast and you cant see animation playing. So what I did is I catch first frame of attack animation and stop animation when it happens again and after that I can switch state.

PR #132, opened 2020-12-07
Golem behavior added. Still not working.

mmacura9:
I am missing something. Putting here for the other team members to see.


source/dagger/gameplay/team_game/golem.cpp
```diff
@@ -0,0 +1,180 @@
...
+		{
+			if (range_.targetFound) golem_.currentAction = EAction::Attacking;
+
+			if (golem_.currentAction == EAction::Idling) {
+				AnimatorPlay(animation_, "ancient_defenders:golem:IDLE");
+			}
+			else if (golem_.currentAction == EAction::Moving) {
+				auto nextPosition = golem_.postition - 1;
+
+				auto coords = WalkingPath::path.back();
```
milosmilosevic13, line 42: You can use auto coords = WalkingPath::path[nextPosition]; since the path will be switched from list to vector.

milosmilosevic13: And then you don't need the loop for finding the element.

source/dagger/gameplay/team_game/golem.cpp
```diff
@@ -0,0 +1,180 @@
...
+				else if (golem_.direction.y == 1) {
+					AnimatorPlay(animation_, "ancient_defenders:golem:ATTACK_FRONT");
+				}
+				else {
+					AnimatorPlay(animation_, "ancient_defenders:golem:ATTACK_FRONT");
+				}
+				Engine::Registry().get<MageStats>(range_.target).health -= golem_.meleeDmg;
+				golem_.currentAction = EAction::Moving; // Go back to moving after attacking
+			}
+			else if (golem_.currentAction == EAction::Chanting) {
```
milosmilosevic13: Golems won't need these 2 states (chanting and defending) so they can be safely removed.

source/dagger/gameplay/team_game/golem.cpp
```diff
@@ -0,0 +1,180 @@
...
+			}
+			else if (golem_.currentAction == EAction::Defending) {
+				// Cast spells
+			}
+		});
+
+}
+
+void ancient_defenders::GolemBehaviorSystem::OnEndOfFrame()
+{
```
milosmilosevic13, line 104: We will move the responsibility of removing characters to HealthManagementSystem (still in works) so this should be removed both here and in mage.cpp.

milosmilosevic13:
Mage expects to find Enemy component of the things it's trying to attack while golem doesn't have that component. GolemStats structure should be renamed to Enemy to be compatible with current code. That is why to game breaks upon contact between characters.

source/dagger/gameplay/team_game/golem.cpp
```diff
@@ -0,0 +1,180 @@
...
+				}
+				else if (transform_.position.y > destinationY) {
+					golem_.direction.y = -1;
+				}
+				else golem_.direction.y = 0;
+
+				if (golem_.direction.x != 0) {
+					sprite_.scale.x = (-golem_.direction.x) * abs(sprite_.scale.x);
+					AnimatorPlay(animation_, "ancient_defenders:golem:WALK_SIDE");
+				}
```
milosmilosevic13: Since the golems don't have different animations for each moving direction it enough to just do AnimatorPlay without checking the directions since the same animation will be played.

source/dagger/gameplay/team_game/golem.cpp
```diff
@@ -0,0 +1,180 @@
...
+						return;
+					}
+				}
+
+			}
+			else if (golem_.currentAction == EAction::Attacking) {
+				if (golem_.direction.x != 0) {
+					sprite_.scale.x = (-golem_.direction.x) * abs(sprite_.scale.x);
+					AnimatorPlay(animation_, "ancient_defenders:golem:ATTACK_FRONT");
+				}
```
milosmilosevic13: Same as for the movement, one AnimatorPlay will be enough.

milosmilosevic13 (approved):
Looks good now. Some things will need to be ironed out in the future but that's the story for another day.

PR #133, opened 2020-12-07
Team/delta snakes/feature/custom collision improvements

UrosBorovcanin:
Made a physics system and implemented a jump feature into the Character Controller FSM


source/dagger/gameplay/team_game/team_game_collisions.cpp
```diff
@@ -0,0 +1,156 @@
...
+    return collisionInfo;
+}
+
+void PlatformerCollisionSystem::LimitPlayerMovement(Collider& collision_)
+{
+    if (!collision_.listOfEntities.empty())
+    {
+        for (int i = 0; i < collision_.listOfEntities.size(); i++)
+        {
+            collision_.canGoRight = (collision_.listOfCollisionSides[i] == CollisionSide::RIGHT) ? false : collision_.canGoRight;//character_.canGoRight;
```
senadkurtisi: We should word wrap these lines to maintain readability

senadkurtisi:
It works really smooth, and I really like the acceleration during falling down

source/dagger/gameplay/team_game/team_game_collisions.h
```diff
@@ -0,0 +1,59 @@
...
+
+        // player, terrain, enemies, traps, collectables, weapons
+        // For player it would be: false, true, true, true, ???, false
+        StaticArray<Bool, 6> collidesWith{ true, true, true, true, true, false };
+
+        CollisionInfo GetCollisionInfo(Vector3& pos_, const Collider& other_, Vector3& posOther_);
+
+    };
+
+    class PlatformerCollisionSystem : public System
```
FedjaVujicic: Might want to rename this and PlatformerCollisionID (line 12) to just CollisionSystem and CollisionID

source/dagger/gameplay/team_game/gravity.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+void team_game::GravitySystem::Run()
+{
+    Engine::Registry().view<Collider, Gravity, Transform>().each([&](Collider& collider_, Gravity& gravity_, Transform& transform_) 
+        {
+            if (collider_.canGoDown)
+            {                
+                if (gravity_.terminalVelocity < m_GravityStrength * gravity_.airTime)
+                {
+                    transform_.position.y -= gravity_.terminalVelocity;
```
FedjaVujicic: You are subtracting velocity [m/s] from a coordinate [m]. I think it's a good idea to change this into gravity_.terminalVelocity * gravity_.airTime, and also the line at 23 to m_GravityStrength * gravity_.airTime * gravity_.airTime and then tweak the JumpSpeed accordingly. It's more realistic that way in my opinion.

source/dagger/gameplay/team_game/gravity.h
```diff
@@ -0,0 +1,25 @@
...
+    struct Gravity
+    {
+        Float32 airTime = 0.f;
+        Float32 terminalVelocity = 5.f;
+    };
+
+    class GravitySystem : public System
+    {
+    private:
+        float m_GravityStrength = 0.5f;
```
FedjaVujicic: Might just be my own preference or a nitpick, but consider renaming this to m_GravitationalConstant or m_GravitationalAcceleration. Makes it easier to understand what it is at first glance :smile:

FedjaVujicic (commented):
Excellent work, works really well in this state. Take a look at these change suggestions and we can merge as soon as you're done.

source/dagger/gameplay/team_game/team_game_collisions.h
UrosBorovcanin: Will do, its a relic from the old system based on the platformer example

source/dagger/gameplay/team_game/gravity.cpp
UrosBorovcanin: Hmm, yeah, that does not seem logical. Also it just occured to me that the actual terminal velocity is framerate dependent which is going to have to be changed

source/dagger/gameplay/team_game/gravity.h
UrosBorovcanin: Yeah, it behaves like an acceleration so i'll change the name accordingly

source/dagger/gameplay/team_game/team_game_collisions.cpp
UrosBorovcanin: And just a question, since this is a non-issue for me, how would you go about wrapping said line?

UrosBorovcanin: I wrapped what i could, but i don't want to wrap further since i think that'll damage the readability in my opinion

source/dagger/gameplay/team_game/team_game_collisions.h
UrosBorovcanin: Renamed

source/dagger/gameplay/team_game/gravity.h
UrosBorovcanin: Renamed

source/dagger/gameplay/team_game/gravity.cpp
UrosBorovcanin: As for this request, I tried making changes requested but i feel it compromises the feel of gravity, which should be linear and not quadratic since i view Transforms to be changes in position over some amount of time(in this case Engine::DeltaTime())

senadkurtisi (approved):

PR #134, opened 2020-12-07
Added collectible system

Aleksa-Jov:
Coins and PowerUps spawn randomly and can be collected


source/dagger/gameplay/tank_warfare/collectibles.cpp
```diff
@@ -0,0 +1,134 @@
...
+				else
+				{
+					if (collectible.collType == ECollectibleType::ExtraLife)
+					{
+						tank.lives++;
+					}
+					if (collectible.collType == ECollectibleType::FireRateUp)
+					{
+						tank.fireRate = 2;
+						tank.powerDuration += 4;
```
DPrivalov: good thing to do here is to make this values into const variables with proper names.

DPrivalov (approved):

source/dagger/gameplay/tank_warfare/collectibles.cpp
Aleksa-Jov: Ok, thanks! I'll do this.

PR #138, opened 2020-12-08
Example/particle system

DPrivalov:
racing game now have simple particles example.

system is placed in separate namespace so every team can easily make a copy and modify it later


DPrivalov:
image

gavrilovmiroslav (approved):

PR #139, opened 2020-12-08
Team/sun tsu/feature/level transition

SofijaBosk:
Created a function for the player to go into different rooms.  Also cleaned up the code for the tilemap entities.


source/dagger/gameplay/team_game/player_controller.cpp
```diff
@@ -10,11 +11,40 @@
...
 
-#include "core/game/transforms.h"
+
 
 using namespace dagger;
 using namespace lab;
 
+void lab::GenerateRoom(int idNext_,lab::NextLvl& currentLvl_, Transform &tr_)
+{
+    if(currentLvl_.id==0)
```
UbiBelETF: This can be moved to data; have a file with lines like:

1 tilemaps/lab/lab.map
2 tilemaps/lab/hallway.map

Load that into a map, and then just send a request from there instead of accumulating branches like this.

source/dagger/gameplay/team_game/player_controller.cpp
```diff
@@ -10,11 +11,40 @@
...
+    {
+        auto& view = Engine::Res<Tilemap>()["tilemaps/lab/lab.map"]->tiles;
+        Engine::Registry().destroy(view.begin(),view.end());
```
DPrivalov: so we are still not cleaning the entire level, why?

DPrivalov: cleaning the whole registry will have less weird bugs

DPrivalov: check out racing example, part of code where we restart the game.

source/dagger/gameplay/team_game/tilemap_entities.cpp
```diff
@@ -0,0 +1,48 @@
...
+        }
+        auto& sprite =reg_.emplace<Sprite>(entity);   
+        if(name_=="create_floor") AssignSprite(sprite,fmt::format("spritesheets:lab:floor_{}", 1 + (rand() % 6)));
+        else AssignSprite(sprite,name_);
+        if(includeCollision_)
+        {
+            auto& col=reg_.emplace<SimpleCollision>(entity);
+            col.size={sprite.size.x/2,sprite.size.y/2};  
+        }
+        sprite.position={ x_ * 16-125.0f, y_ * 16-b_, depth_ };
```
DPrivalov: sprite position will be set from transform later. No need to set it here.

DPrivalov: trn.position={ x_ * 16-125.0f, y_ * 16-b_, depth_ };

DPrivalov (changes requested):

source/dagger/gameplay/team_game/player_controller.cpp
SofijaBosk: I thought not to clear the whole level so the player stats (ex health from room to room) don't change. Otherwise we would have to backup everything then put another player entity.

TeodorVasiljevic365 (approved):

PR #141, opened 2020-12-09
Team/delta snakes/feature/add collision to platforms

FedjaVujicic:
Added collisions to platforms. Improved the gravity system and made it frame rate independent.


UrosBorovcanin (approved):
As we have alreday discussed these imolementations and fixes on discord, i think it works very well and has a really good basis for the rest of our game

senadkurtisi (approved):
It works really well now. The jumping is not floaty anymore, and most importantly, it's not framerate dependent.

PR #142, opened 2020-12-09
Added health management, and HP bars for characters.

milosmilosevic13:
Health is now a separate component. Health management system takes care of displaying HP bars under characters as well as removing characters whose HP drops below zero.


source/dagger/gameplay/team_game/hp_system.cpp
```diff
@@ -0,0 +1,81 @@
+#include "hp_system.h"
+
+#include "core/engine.h"
+#include "core/graphics/sprite.h"
+#include "core/game/transforms.h"
+
+#include <iostream>
```
mmacura9, line 7: I think you do not use this.

source/dagger/gameplay/team_game/mage.cpp
```diff
@@ -103,7 +105,7 @@ void ancient_defenders::MageBehaviorSystem::Run()
             else {
                 AnimatorPlay(animation_, "ancient_defenders:mage:ATTACK_FRONT");
             }
-            Engine::Registry().get<Enemy>(range_.target).health -= mage_.meleeDmg;
+            Engine::Registry().get<Health>(range_.target).currentHealth -= mage_.meleeDmg * Engine::DeltaTime();
```
mmacura9, line 109: This should be also added to golem.h/ golem.cpp

source/dagger/gameplay/team_game/hp_system.cpp
milosmilosevic13: You are right, it's a leftover from some testing, will be removed.

source/dagger/gameplay/team_game/mage.cpp
mmacura9: I mean Health should be added to golem

mmacura9 (approved):
Everything is fine, we just should add this to golem.

PR #145, opened 2020-12-10
Added camera following system

senadkurtisi:


FedjaVujicic (approved):
Good job. In the future we might want something a bit more dynamic, but this is a nice prototype for now and will allow us to more easily work on other aspects of the game.

senadkurtisi:
> Good job. In the future we might want something a bit more dynamic, but this is a nice prototype for now and will allow us to more easily work on other aspects of the game.

I think that maybe a camera with some sort of latency would be a nice thing to have. But not something with a hard start and hard stop latency. Maybe a soft start camera which tries to "catch up" with the player but in a smooth way.

What is your opinion on this? @UrosBorovcanin

UrosBorovcanin (approved):
For now this camera follow system is completely fine. But i do agree with the general consensus. If the time allows us we should look into making a smoother camera system since this one does feel a bit too rigid.

PR #146, opened 2020-12-11
Tower system implemented

milosmilosevic13:
Added a tower system that takes care of building towers by checking if they are done, and attacking enemies that are close. Also added the Chanting mage state that is responsible for increasing the towers' building progress.


source/dagger/gameplay/team_game/towers.cpp
```diff
@@ -0,0 +1,88 @@
...
+            sprite.pivot = { 0,0.25f };
+            sprite.scale = { 2,2 };
+        }
+
+        if (range_.targetFound) tower_.currentAction = EAction::Attacking;
+
+        AnimatorPlay(animator_, "ancient_defenders:towers:" + tower_.type);
+        if (tower_.currentAction == EAction::Idling) {}
+        else if (tower_.currentAction == EAction::Attacking) {
+            Engine::Registry().get<Health>(range_.target).currentHealth -= 15.0f*Engine::DeltaTime();
```
DPrivalov, line 46: 15.0f should be some const or variable

source/dagger/gameplay/team_game/towers.h
```diff
@@ -0,0 +1,59 @@
...
+        static Float32 constructionGoal;
+
+        EAction currentAction = EAction::Idling;
+        String type;
+    };
+
+
+    struct Tower {
+        Entity entity;
+        Transform & coordinates;
```
DPrivalov, line 32: don't save components, get them from entity every frame.

DPrivalov (changes requested):

source/dagger/gameplay/team_game/towers.cpp
milosmilosevic13: This part of the code will be re written, this is just a placeholder for testing things.

milosmilosevic13:
A lot of this code was changed when controls got added, due to a number of bugs that have appeared; this will be added along with the controls and a bunch of fixes together in a new pull request.

PR #150, opened 2020-12-12
Team/sun tsu/feature/enemy movement

TeodorVasiljevic365:


source/dagger/gameplay/team_game/enemy.cpp
```diff
@@ -22,42 +22,88 @@ void lab::EnemySystem::Run()
...
+		Vector2 playerPosition;
+		auto view3 = reg.view<Sprite, Animator, InputReceiver, Transform, PlayerCharacter, SimpleCollision>();
+		for (auto entity : view3)
+		{
+			auto playerTransform = view3.get<Transform>(entity);
+			playerPosition = playerTransform.position;
+			playerPosition.y *= -1;
+		}
+
+		if (bandit.ID == 1)
```
DPrivalov: could you use Enum instead of ID here? it's a bit hard to read code right now

DPrivalov (approved):

source/dagger/gameplay/team_game/enemy.cpp
TeodorVasiljevic365: I agree, I will add Enums

PR #153, opened 2020-12-17
team/morning_afternoon/feature/Dash

Coakim:
Simple dash forward when double tapping movement key whitch drains stamina

### DIFF
--- a/source/dagger/gameplay/plight/plight_combat.cpp
+++ b/source/dagger/gameplay/plight/plight_combat.cpp
@@ -62,7 +62,7 @@ void PlightCombatSystem::Run()
 
 			//Stamina will not be affecting running , it will be used for special movement like dashing or rolling when it gets implemented (used on running for example here)
 
-			if (character.running) {
+			/*if (character.running) {
 
 				cstats.currentStamina -= STAMINA_FOR_RUNNING_FRAME;
 				if (cstats.currentStamina < STAMINA_FOR_RUNNING_FRAME) {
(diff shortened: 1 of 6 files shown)


source/dagger/gameplay/plight/plight_controller_fsm.cpp
```diff
@@ -25,7 +25,7 @@ void PlightCharacterControllerFSM::Idle::Enter(PlightCharacterControllerFSM::Sta
 	auto& animator = Engine::Registry().get<Animator>(state_.entity);
 	AnimatorPlay(animator, "Plight:big_deamon:IDLE");
     auto& character = Engine::Registry().get<PlightCharacterController>(state_.entity);
-    character.running = false;
+    character.running = false; character.dashing = false;
```
DPrivalov, line 28: why is it on the same line?

source/dagger/gameplay/plight/plight_combat.cpp
```diff
@@ -62,7 +62,7 @@ void PlightCombatSystem::Run()
 
 			//Stamina will not be affecting running , it will be used for special movement like dashing or rolling when it gets implemented (used on running for example here)
 
-			if (character.running) {
+			/*if (character.running) {
```
DPrivalov, line 65: do u need running still?

DPrivalov (approved):

source/dagger/gameplay/plight/plight_controller_fsm.cpp
Coakim: Accident probably ...

source/dagger/gameplay/plight/plight_combat.cpp
Coakim: I left it commented in case we need it during development for some reason , or if we decide to add cost for running as 
well

PR #154, opened 2020-12-17
Added Tower and Controls Systems along with some fixes

milosmilosevic13:
Implemented Towers and added player controls. Also a lot of changes happened to fix some of the issues that started happening once all these systems were put together. Removed some outdated code that was either just commented out, or unused.


source/dagger/gameplay/team_game/controls.cpp
```diff
@@ -0,0 +1,74 @@
...
+#include "mage.h"
+#include "golem.h"
+#include "towers.h"
+
+#include "core/input/inputs.h"
+#include "core/engine.h"
+
+using namespace ancient_defenders;
+
+Float32 ancient_defenders::ControlData::timeout = 0.0f;
```
DPrivalov: it can be a simple struct member inside PlayerControlsSystem. No need to make it static.

maxTimeout  can be a constexpr value.

source/dagger/gameplay/team_game/golem.cpp
```diff
@@ -39,18 +38,11 @@ void ancient_defenders::GolemBehaviorSystem::Run()
...
-						break;
-					}
-					i++;
-				}
+				auto coords = WalkingPath::path[nextPosition];
+				
 
-				auto destinationX = coords.x; // Maybe add +/- a few percent to make paths bit more varied
-				auto destinationY = coords.y;
+				auto destinationX = coords.x + golem_.offset.x;
```
DPrivalov: might be good to rename golem_ to enemy_ too

source/dagger/gameplay/team_game/golem.cpp
```diff
@@ -93,66 +88,65 @@ void ancient_defenders::GolemBehaviorSystem::Run()
...
 
-	AssignSprite(gol.sprite, "spritesheets:golem-little-sheet:golem_stand_side:1");
-	float ratio = gol.sprite.size.y / gol.sprite.size.x;
-	gol.sprite.scale = { 4,4 };
+    // Randomly create offset 
+    std::random_device dev;
+    std::mt19937 rng(dev());
 
-	auto start = WalkingPath::path.back();
+    std::uniform_int_distribution<std::mt19937::result_type> roll22(0, 22);
```
DPrivalov, line 124: 22, 38 and other magic numbers can be made into constexpr values, that way u will get the same compilation time but better readability.

source/dagger/gameplay/team_game/golem.cpp
```diff
@@ -93,66 +88,65 @@ void ancient_defenders::GolemBehaviorSystem::Run()
...
-Golem ancient_defenders::Golem::Create()
+Entity ancient_defenders::Golem::Create()
 {
 	auto& reg = Engine::Registry();
 	auto entity = reg.create();
-	auto gol = Golem::Get(entity);
+	auto& sprite = reg.emplace<Sprite>(entity);
+	auto& coordinates = reg.emplace<Transform>(entity);
+	auto& anim = reg.emplace<Animator>(entity);
+	auto& golem = reg.emplace<Enemy>(entity);
```
DPrivalov, line 111: golem to enemy here as well

source/dagger/gameplay/team_game/towers.h
```diff
@@ -0,0 +1,62 @@
...
+using namespace dagger;
+
+namespace ancient_defenders {
+
+    enum class SelectMode {
+        Spot = 1,
+        Tower = 2,
+    };
+
+    struct TowerPlacementInfo {
```
DPrivalov, line 17: @gavrilovmiroslav, take a look at this code. can u suggest a better way of doing this?

DPrivalov (approved):
minor things and good to go

source/dagger/gameplay/team_game/golem.cpp
milosmilosevic13: Forgot to change it here as well, it will be changed.

source/dagger/gameplay/team_game/golem.cpp
milosmilosevic13: Actually here the name can stay golem, since this is Golem::Create function. We had in mind adding different enemies that would have their own Create functions due to the small differences in their data/stats/assets, so leaving the name unchanged here might provide better code readability later when more enemy types are in play.

source/dagger/gameplay/team_game/towers.cpp
```diff
@@ -0,0 +1,86 @@
+#include "towers.h"
+#include "mage.h"
+
+#include "core/engine.h"
+
+using namespace ancient_defenders;
+
+UInt32 ancient_defenders::TowerPlacementInfo::selectedSpot = 0;
```
gavrilovmiroslav, line 8: For later, this definitely needs to change to being loaded from data and not hardcoded, but for now leave it and get the game up and running.

source/dagger/gameplay/team_game/towers.h
gavrilovmiroslav: Agreed, there shouldn't be a struct full of static things anywhere :D 

We'll talk about it when all the features are working, how does that sound, team?

milosmilosevic13: Sound good, this was the easiest way to get access to and modify all this data from a bunch of different spots.

nkristina (approved):

PR #156, opened 2020-12-18
namespaces issue fixed, linker error are still in place

DPrivalov:
- error LNK2001: unresolved external symbol "public: virtual void __cdecl platformer::PlatformerCharacterControllerFSM::Idle::Enter(struct FSM<enum platformer::EPlatformerCharacterStates,void>::StateComponent &)" (?Enter@Idle@PlatformerCharacterControllerFSM@platformer@@UEAAXAEAUStateComponent@?$FSM@W4EPlatformerCharacterStates@platformer@@X@@@Z)

### DIFF
--- a/source/dagger/gameplay/platformer/character_controller_fsm.cpp
+++ b/source/dagger/gameplay/platformer/character_controller_fsm.cpp
@@ -7,42 +7,41 @@
 #include "gameplay/platformer/platformer_controller.h"
 
 using namespace dagger;
+using namespace platformer;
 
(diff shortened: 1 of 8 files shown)


gavrilovmiroslav:
I don't think we should merge this before the linker errors are gone too. I'm struggling to understand them, but vitally, they are VS-related (or rather sharpmake-related, I'm afraid). The way the linker gets the files may somehow be broken, @rtojagic might need to get involved. :)

PR #157, opened 2020-12-18
Team/morning afternoon/feature/projectile system

Coakim:
Implemented a basic projectile system based on the example particle system. Each character has a projectile spawner which spawns a new projectile in the direction of the crosshair on a keybind release


source/dagger/gameplay/plight/plight_projectiles.cpp
```diff
@@ -0,0 +1,109 @@
...
+}
+
+void ProjectileSystem::WindDown()
+{
+    Engine::Dispatcher().sink<NextFrame>().disconnect<&ProjectileSystem::OnEndOfFrame>(this);
+}
+
+void ProjectileSystem::OnEndOfFrame()
+{
+   //Need to change this part  to check if projectile colided with something , if yes destroy it
```
DPrivalov, line 96: you can mark it as toDestroy=true in some damagesystem or something after it collided and damage was processed properly. and then remove it here.

DPrivalov (approved):

source/dagger/gameplay/plight/plight_projectiles.cpp
Coakim: Nice idea ... I was thinking about something similar , but waiting for team to finish collision system to put that part in

PR #158, opened 2020-12-18
Tilemap system

Yugo0:
So, this branch is kind of a mess, beacuse it was started a few weeks ago, and in meantime I had to pull from both `main` and `team/sultans_of_swing/main`. It's tilemap system with a basic level example that Smiljana and I made.
I am making this pull request now, because I need this to start working on optimizing physics system.


kostaDimitrijevic (approved):

source/dagger/gameplay/team_game/level_generator.h
```diff
@@ -0,0 +1,36 @@
+#pragma once
+
+#include "core/core.h"
+
+namespace level_generator
+{
+    namespace smiljana
```
DPrivalov, line 7: a bit weird to split code like that, hope you will merge with more meaningful namespaces later. 
You can leave names in comments thou.

DPrivalov (approved):

source/dagger/gameplay/team_game/level_generator.h
Yugo0: Yes, we organized it like this so each of us could build their levels and test them and not worry too much about function names. These inner namespaces will be removed and we will settle on singular set of functions in the end.

smiljanaspasic (approved):

PR #160, opened 2020-12-18
Detection system

kostaDimitrijevic:
I added detection system that uses circular collision


DPrivalov (approved):
looks fine, 
need to see it's usage

PR #161, opened 2020-12-18
Fix for character controller 

kostaDimitrijevic:
Made few changes for character controller. I did changes because I needed information on current shape form of our hero because of detection. Before I needed to use CHaracterControllerSystem, this way was easier.


source/dagger/gameplay/team_game/character_controller.cpp
```diff
@@ -21,17 +21,20 @@ using namespace team_game;
 
 void CharacterControllerSystem::Run()
 {
-	Engine::Registry().view<CharacterFSM::StateComponent>().each(
-		[&](CharacterFSM::StateComponent& state_)
-		{   
-		auto& input = Engine::Registry().get<InputReceiver>(state_.entity);
-		if (input.Get("goblinTransform") == 1) { idle = "among_them_animations:goblin_idle"; running = "among_them_animations:goblin_run"; SetShape(ECharacterShape::Goblin); }
-		if (input.Get("slimeTransform") == 1) { idle = "among_them_animations:slime_idle"; running = "among_them_animations:slime_run"; SetShape(ECharacterShape::Slime);		}
-		if (input.Get("batTransform") == 1) { idle = "among_them_animations:bat"; running = "among_them_animations:bat"; SetShape(ECharacterShape::Bat); }
```
DPrivalov, line 30: SetShape was better =) u could add logs easier and other logic as well

DPrivalov (approved):
more info about what was fixed and how would be good here

source/dagger/gameplay/team_game/character_controller.cpp
kostaDimitrijevic: I changed it because I needed information on current shape form of hero for detection. It's easier to access like this rather then information to be private in CharacterControllerSystem. I would need an object to use get and set.

PR #164, opened 2020-12-18
skeleton death animations added

TeodorVasiljevic365:


source/dagger/gameplay/team_game/enemy.h
```diff
@@ -17,7 +17,7 @@ namespace lab
 		int cooldown = 120;
 		int maxCooldown = 120;
 		int health = 100;
-		Float32 timeWounded = 100;
+		Float32 deathTimer = 125;
```
DPrivalov, line 20: I think it's better to do it in msec or some time related value, so it will work properly on all machines (with different fps)

DPrivalov (approved):

PR #165, opened 2020-12-18
Team/delta snakes/feature/adding grounded check

UrosBorovcanin:
Instead of adding a grounded check this branch fixes the already existing canGoX set of flags to do what they are supposed to. On another note, when it comes to merging Fedja might want to take a look into this since it has some changes to how gravity functions and disables Terminal Velocity

### DIFF
--- a/source/dagger/gameplay/team_game/team_game_collisions.h
+++ b/source/dagger/gameplay/team_game/team_game_collisions.h
@@ -35,6 +35,8 @@ namespace team_game
         Bool canGoUp{ true };
         Bool canGoDown{ true };
 
+        Bool grounded = false;
+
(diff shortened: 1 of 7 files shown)


source/dagger/gameplay/team_game/team_game_collisions.h
```diff
@@ -35,6 +35,8 @@ namespace team_game
         Bool canGoUp{ true };
         Bool canGoDown{ true };
 
+        Bool grounded = false;
```
DPrivalov: don't see if u use it anywhere

UrosBorovcanin: Oh, right, that's a remnant of an older idea, will be removed

senadkurtisi (approved):

FedjaVujicic (approved):

PR #171, opened 2020-12-19
engine/fix/unmapped_keys

milosmilosevic13:
Simple fix for the issue that caused engine to crash if an unmapped keyboard button was pressed.


source/dagger/core/input/inputs.cpp
```diff
@@ -14,8 +14,10 @@ using namespace dagger;
 
 void InputSystem::OnKeyboardEvent(KeyboardEvent input_)
 {
+    if ((SInt32)input_.key < 0) return;
```
DPrivalov: try to use {} even for one line, it's more readable that way 

{
    return;
}

DPrivalov (approved):

PR #173, opened 2020-12-19
Team/sultans of swing/feature/patrolling enemies

smiljanaspasic:
Ok,so this solution works for now but I see one huge flaw regarding the FileInputStream inFile{ "path.txt" }; which I put in Enemy.cpp since we will have different types of enemies and they will all need different paths,so I'll try to think of a solution for this in the meantime
EDIT: Fixed


Yugo0 (approved):

smiljanaspasic:
Check it out and if everything is fine I'll merge it

kostaDimitrijevic (approved):

PR #180, opened 2020-12-20
Team/morning afternoon/feature/collison

DamjanTM:
Implemented the collision & physics systems.

(Note: I think the changes to window.cpp are form pulling a previous main into this branch)


source/dagger/gameplay/plight/plight_physics.cpp
```diff
@@ -0,0 +1,71 @@
...
+                    for (int group : other_physics.collision_groups) {
+                        if (std::find(physics.my_groups.begin(), physics.my_groups.end(), group) != physics.my_groups.end()) {
+                            has = true;
+                            break;
+                        }
+                    }
+                    if (has) {
+                        auto& other_transform = view.get<Transform>(*it2);
+                        auto& other_collision = view.get<PlightCollision>(*it2);
+                        ResolveCollision(transform.position, collision, *it, other_transform.position, other_collision, it2);
```
Coakim: In case the two non static physics objects colide this always resolves the collision as if the first physics object that the iteration loop checks initiated the collision.This can be seen when testing the collision by moving one character and colliding with the other , and then moving the other character and colliding with the first one.In the first occasion the collision is resolved correctly as the character which is moving is dealt with the first in the loop and he is moved back properly.But the problem is when you move the other character into the first character , the physics system checks the collision for the character which is currently not moving and then resolves the collision as if he was the one moving , so basically the other character is pushing the static character back instead of him being pushed back from the collision.I think this can be resolved easily by also checking if the character whose PhysicsObject component we are checking is also in running or dashing state which can be found out of the PlightCharacterController struct with the same entity as the PhysicsObject and then just not resolving collision if he is not actually running or dashing because then he can't be the one initializing the collision

Coakim (changes requested):

Coakim (approved):

PR #187, opened 2020-12-21
Poof effect on shape change

Yugo0:
In order to make this i had to change `Animator` component and `AnimationSystem`, so they can play animation only once instead of looping it indefinitely.


smiljanaspasic (commented):
Nice job! I actually had another idea for implementing this but yours is far better 🥇

smiljanaspasic (approved):

PR #200, opened 2020-12-22
Added icons and mouse support to the game

gavrilovmiroslav:
Spot selection is now done by mouse hover alone. Tower selection is followed with UI elements when spot is hovered. Click to place tower-building mage. Tower building process shows which tower is being built and how many mages are building.


source/dagger/gameplay/team_game/controls.cpp
```diff
@@ -3,69 +3,155 @@
...
+
+                found = true;
+                break;
+            }
+        }
+    }
+
+    if (!found)
+    {
+        menu.position = choice.position = build.position = buildIconTransform.position = { 10000, 10000, 10000 };
```
milosmilosevic13, line 123: Circle.position should be changed here as well, since right now it stays in place after mouse cursor hovers over a spot.

milosmilosevic13 (approved):
Looks great!

mmacura9 (approved):

PR #203, opened 2020-12-22
Level reloading

Yugo0:
It's now possible to reload level with R key.


source/dagger/gameplay/team_game/game_controller.h
```diff
@@ -0,0 +1,20 @@
...
+#include "core/system.h"
+
+using namespace dagger;
+
+namespace team_game
+{
+	struct GameController
+	{
+		UInt32 id{ 0 };
+	};
```
smiljanaspasic, line 12: I don't think this id here is really necessary

source/dagger/gameplay/team_game/team_game_main.cpp
```diff
@@ -144,104 +146,109 @@ void SetupWorldSmiljana(Engine& engine_, Registry& reg_) {
...
+	auto& map = reg.emplace<StaticBodyMap>(mapEnt);
+	Engine::PutDefaultResource<StaticBodyMap>(&map);
 	
     // You can add your own WorldSetup functions when testing, call them here and comment out mine
-    //SetupWorldJovica(engine_, reg);
+    SetupWorldJovica(engine_, reg);
     //SetupWorldKosta(engine_, reg);
-    SetupWorldSmiljana(engine_);
+    //SetupWorldSmiljana(engine_, reg);
 }
```
smiljanaspasic: Try putting here that id as a global parameter and set it to 0 outside of SetupWorld and then try something like 
if(id==0) {
SetupWorldJovica(engine_, reg); }
if(id==1) {  
SetupWorldSmiljana(engine_, reg); }
id++;
if(id==2) {id=0;}

smiljanaspasic (changes requested):
The reason why I'm suggesting this changes now is because I think that floor tile that appears when you reload the level isn't so naive as it seems.I suspect there's more to it,first something else also appears whenever I try to reload just SetupWorldSmiljana(while SetupWorldJovica is commented),also when I tried to implement this changes myself to switch between different SetupWorlds I had to press R multiple times to switch between mine SetupWorld and yours
Maybe you could look a little bit more into this,it seems to me like the Registry isn't fully cleared somehow but I'm not sure why

smiljanaspasic (approved):
This solved everything 👍

PR #205, opened 2020-12-22
Added collectables system.

senadkurtisi:
Simple collectables system. Speed up, and health are implemented. New texture loading system for collectables specifically is implemented.


FedjaVujicic (approved):
Good job. As we discussed, we probably won't be needing a health system for the game currently, but the speed collectible works nicely, and the position and duration are exactly right.

UrosBorovcanin (approved):
This addition works well, and can be used to great extent. As for the health system i don't think it should be removed just in case we miraculously implement it at some point in time

PR #230, opened 2020-12-25
Team/delta snakes/fix/collision performance

FedjaVujicic:
Greatly improved the performance of collisions


UrosBorovcanin (approved):
The code works really well and has bumped up the performance by quite a bit. Nice work!

PR #235, opened 2020-12-25
Team/morning afternoon/feature/defense field

Coakim:
Added an ability for the player to create a force field around him that acts as a shield from damage . While the player is in the field he drains stamina but also regenerates health. After the shield disappears the player can then regenerate stamina normally and continue to play the game. Each player has 2 defensive fields per game round . Also , the other player can't go in another players field


DamjanTM (approved):
Works great, except when it moves the player out of bounds

PR #236, opened 2020-12-25
Team/delta snakes/fix/crash on hitting trap

UrosBorovcanin:
The game crashed in camera_follow.cpp on line 42 because of an empty iterator


senadkurtisi (approved):
Good job. This conflict happened due to the removal of the player entities from the registry when the end game condition is met.

PR #237, opened 2020-12-25
Team/delta snakes/feature/distance loss condition

UrosBorovcanin:
The player who goes out of bounds now loses, and on restart the camera goes back to {0, 0} which solves some reset bugs


senadkurtisi (approved):
Excellet job. This completes our end game screen functionality.

PR #243, opened 2020-12-25
Team/delta snakes/feature/level creation

FedjaVujicic:
Finished level creation.


UrosBorovcanin (approved):
A pretty nice level with examples of both jump and speed boost pickups. Good work!

